PR #2, opened 2022-07-04
Added react vite + react-query example

paul-sachs:


akshayjshah:
Wait, are we not planning to include a react-query example?

timostamm:
We are actually planning a more direct integration with React Query that gives you a ReactQueryClient - similar to the [PromiseClient and CallbackClient](https://connect.build/docs/web/using-clients) we already provide. 

This was only closed because it is not ready yet, but it will resurface soon.

paul-sachs:
Yep, I just want it to bake a little internally before putting out examples using it. The plan is to actually use it within the BSR/Studio. Gonna dogfood it.

akshayjshah:
Awesome!

PR #6, opened 2022-08-05
add vite example

utamori:
Thanks for a great library!
I tried it with vite and react. I'll add it as it doesn't seem to exist in the samples.


react/vite/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
```
smaye81, line 1: Most of these are duplicated in the main `.gitignore` in the project root.  Can you consolidate all of these to that file instead?

react/vite/README.md
```diff
@@ -0,0 +1,35 @@
+# Eliza
+
+This project was bootstrapped with [Vite](https://github.com/vitejs/vite) using the command:
+
+`npm create vite vite --template react-ts`
+
+All default tooling setup via Vite is used, which includes:
+
+-   [Vite](https://ja.vitejs.dev/) (Build Tool)
```
smaye81, line 9: I think we can make this comment just say:

> All default tooling setup via the above command is used.

Probably self-explanatory that Vite itself is used.

react/vite/README.md
```diff
@@ -0,0 +1,35 @@
...
+-   [Vitest](https://vitest.dev/) (Test Framework)
+
+## Getting Started
+
+### `npm install`
+
+### `npm start`
+
+Runs the app in the development mode.\
+Open [http://localhost:5173](http://localhost:5173) to view it in the browser.
```
smaye81, line 19: All the sample apps using port `3000` as their default port.  Can you change this to run there instead?

react/vite/src/App.test.tsx
```diff
@@ -0,0 +1,26 @@
+import {
```
smaye81, line 1: If I run the unit tests, I'm seeing failures related to the Fetch API.  I think it's just a matter of how things are configured.  If you look at our Vue examples, you should see an example for passing tests using Vitest.

smaye81 (changes requested):

smaye81:
Thanks for the PR!  I left a few comments, but I saw you just closed it.  Either way, appreciate the interest.  Thanks again!

smaye81 (approved):

PR #8, opened 2022-08-06
Adding new react-query example.

paul-sachs:
Experimenting with a react-query integration.

API still in progress.


react/react-query/src/create-query-client.tsx
```diff
@@ -0,0 +1,213 @@
...
+
+type OneofSelectedMessage<K extends string, M extends Message> = {
+    case: K
+    value: M
+}
+
+// Made a recursive version of PlainMessage which plays better with 
+type PlainMessage<T extends Message> = {
+  [P in keyof T as T[P] extends Function ? never : P]: PlainMessageField<T[P]>;
+};
```
paul-sachs, line 44: @timostamm I wrote a recursive PlainMessage here. I wasn't sure if I wanted to kick this back upstream to @bufbuild/protobuf-es or just leave it here (since it's largely just helpful to use for react state updates).

Thoughts?

paul-sachs: Most of the time, the default `PlainMessage` is fine, but this Service I'm hitting has some responses that contain other messages which need to be cleaned up.

timostamm: Nice! For now, it is best kept here, where we can gain more experience with it. Could you elaborate a bit on what was missing from @bufbuild/protobuf's `PlainMessage`? I'm super interested to know what the issue was.

paul-sachs: I ran into an issue while trying to mutate the stored data in a query [here](https://github.com/bufbuild/connect-web-integration/pull/8/files#diff-5ba2c2331db0f4296e24d8d6d9f9872305a1933e44cc4fff5ee4d7ddef90ce89R177-R196). Basically, as a response I get `GetTopRepositoriesResponse` which is stripped of extra props via the original PlainMessage, but within that entity is an array of `TopRepository`, which is not stripped of those class properties. If I try to do a spread with that object, my result obv is missing those class properties and typescript complains because the returned object does not meet the requirements.

smaye81: @timostamm How does this affect https://linear.app/bufbuild/issue/TCN-263/fix-plainmessaget?  Should we still revert that back to its original and then leave this here to let it bake for a bit?

For context, TCN-263 would revert back to:

```js
export type PlainMessage<T extends Message> = {
  // eslint-disable-next-line @typescript-eslint/ban-types -- we use `Function` to identify methods
  [P in keyof T as T[P] extends Function ? never : P]: T[P];
};
```

timostamm: > I ran into an issue while trying to mutate the stored data in a query

I suppose `setQueriesData()` takes a `PlainMessage` as an argument? Wouldn't it be desirable to have it take a `PartialMessage` instead?

paul-sachs: > I suppose `setQueriesData()` takes a `PlainMessage` as an argument? Wouldn't it be desirable to have it take a `PartialMessage` instead?

That works a little better than just the original PlainMessage but since an updater must take and return the same data structure, it would allow us to remove props where we shouldn't, and also would require way more `?.` operators to make accessing type safe.

paul-sachs: For a little more context, it doesn't take a PlainMessage, it takes a callback that receives the previous PlainMessage and returns a new one.

timostamm: Is there a reason we can't relax this rule about the updater, and give the user `T`, but accept a `PlainMessage<T>` or `PartialMessage<T>` in return? We can always normalize with `new T()` (not exactly, but the constructor should be available unless I'm missing something).

paul-sachs: Hmm, I think I could actually new something with that yeah. I was trying to fit our structures more closely with the ones I'm used to (ie, plain objects) rather than moving the other direction. Storing class instance in our objects would work, though we'd have to figure out some custom serialization/deserialization for integrating things like [persistQueryClient](https://tanstack.com/query/v4/docs/plugins/persistQueryClient).

As much as I think calling `new T()` might make sense for our specific case, the React ecosystem is more aligned around storing plain objects.

paul-sachs: We could definitely write a persister that works with our message format (detect that the stored result is a message type and serialize it using a custom serializer) but it does feel a little like we're fighting against the platform.

timostamm: I'm not familiar with TanStack Query persisters, but if we are going to persist _any_ part of a protobuf message, we must not `JSON.stringify` it. As explained earlier, a protobuf message can contain types such as a `BigInt`, but also `NaN`, `Infinity` and `-Infinity`. Plain JSON cannot represent those values and will crash or lose data. The proper way is to use the messages `toJson` method (which implements the [proto3 JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json)), or `toBinary` - which is generally preferable for performance. 

Does that mean we do need to implement a persister?

Regarding `PartialMessage<T>`/ `PlainMessage<T>`/`T`, it is important to be aware of two other characteristics of protobuf:

1) Adding a field to a message should not break application code. This is why we have constructors. They take a `PartialMessage` because that happens to align with most use cases, letting users only specify fields they intend to provide a value for. Scalar / map / repeated fields are all optional on the wire, but they need to be non-nullable types in the runtime representation (with scalar fields it is a must, because of intrinsic default values in proto3 - `int32 field = 1` must have an initial value `0`). 

2) Only the concrete message type provides the field information required for important features: 
  a) JSON and binary serialization, 
  b) fast, deep equality checks with `equals()`, 
  c) deep clone with `clone()`.
  It is also very convenient to have class names appear in stack traces and console logs, especially with deeply nested structures. This is why I think that in general, API surfaces should consume partial or plain input, but produce concrete types. 

This is what prompted me to suggest we should try to follow the pattern explained in 2) with `setQueriesData`. Peeking at the [source](https://github.com/TanStack/query/blob/52acd7892969fc726f10d4141f63ee7f0c6ae4c2/packages/query-core/src/utils.ts#L65
), it seems we could easily do that with our `createQueryUpdater()`?

paul-sachs: Right right, so if we ever wanted to support these persistence adapters, we couldn't rely on the defaults one way or the other. That makes more sense. 

Using the contractor is easy, I can easy update our helper to automatically do that and probably will.

paul-sachs: We don't NEED to implement a persister, I was mostly just trying to think of situations which may cause some friction within the react-query ecosystem.

The persistence layer is pretty easy to write and I might do an example one here for reference though.

paul-sachs: @timostamm updated code and cleaned up the structure a little bit. The library now actually has no explicit dependency on react-query itself, only a dependency on `react`, `@bufbuild/protobuf-es` and `@bufbuild/connect-web`.

react/react-query/README.md
```diff
@@ -0,0 +1,7 @@
+Things to note
+- [] Default hash function doesn't support BigInt, we're using stable-hash as an alternative.
+- [x] Better define provider and query file structure. Split out code that requires react-query as an import
+- [x] Consider and alternative that provides config directly instead of wraping react-query
+- [] Is there any benefit to automatically invalidating queries after mutation? Without being able to connect APIs with a relationship this might actually be a moot point.
```
tannerlinsley, line 5: This is precisely why RQ doesn't do it automatically. It would require schema which really only comes when you have access to the database structure (graphql or manual). We opted to make the user opt-in to mutations and it's been pretty nice TBH. Usually they are only invalidating a whole set of queries like so: `queryClient.invalidateQueries(['todos'])` which would invalidate any queries with `todos` in the first array key slot (eg. `['todos']` and `['todos', 5]`.

react/react-query/src/react-query-connect-web/use-query-hooks.tsx
```diff
@@ -0,0 +1,43 @@
...
+import {
+    QueryClient,
+    createQueryClient,
+} from './create-query-client'
+import { ServiceType } from '@bufbuild/protobuf'
+import { createConnectTransport, Transport } from '@bufbuild/connect-web'
+
+const hooksContext = createContext(
+    new Map<ServiceType, QueryClient<ServiceType>>()
+)
```
tannerlinsley, line 13: I'm curious why this is a context. Future proofing for dep injection?

paul-sachs: This was really just a fast and loose way of making sure we only ever created a single version of a client per service. Makes sure our hooks are stable. I suppose we could just use a map stored at the module level.

react/react-query/README.md
```diff
@@ -0,0 +1,8 @@
+Things to note
+- [] Default hash function doesn't support BigInt, we're using stable-hash as an alternative.
+- [x] Better define provider and query file structure. Split out code that requires react-query as an import
+- [x] Consider and alternative that provides config directly instead of wraping react-query
+- [] Is there any benefit to automatically invalidating queries after mutation? Without being able to connect APIs with a relationship this might actually be a moot point. 
+- [x] Updating cached state feels clunky. This api needs some work.
+- [] Since we are now storing class objects in the cache in order to preserve advanced protobuf functionality, we'll need some examples of a persister
+- [] Tried using a custom hash function on a per query basis (thus allowing more fine grained control of hashes specific to connect-web) but this prevents proper usage of `setQueryData` since those functions do not accept a hash function.
```
paul-sachs, line 8: @tannerlinsley I was noodling on this question with @timostamm. Would you be amenable for a PR against react-query to add an options object to `setQueryData` and related functions to accept a custom hash key function?

paul-sachs:
Closing since we have our own dedicated repo for connect-query.

PR #11, opened 2022-08-25
Add an example using plain JS

### LINKED ISSUE #10: add a plain javascript example(ie, no frameworks)
Is it possible to use `connect-web` with just javascript without any framework? If yes, I would appreciate an example been added to this repo.

Thanks.

smaye81:
This adds an example using just plain JS (no frameworks).  The only dependencies are:

- Protobuf-ES
- Connect-Web
- TypeScript
- esBuild

Fixes #10


plain/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+document.getElementById("user-input")?.addEventListener("keyup", (event) => {
+    event.preventDefault();
+    if (event.key === "Enter") {
+        document.getElementById("send-button")?.click();
+    }
+});
+
+// Adds a node to the DOM representing the conversation with Eliza
+function addNode(text: string, sender: string): void {
+    const respContainerEl = containerEl.appendChild(document.createElement('div'));
```
timostamm, line 32: It would be best practice to construct elements first, then add them to the DOM. Otherwise, the browser needs to recalculate the layout several times. It will not make a noticeable difference here, though...

timostamm (approved):

plain/src/index.ts
smaye81: Duly noted.  I updated it anyway.

PR #15, opened 2022-09-17
Update Svelte example

### LINKED ISSUE #9: Update sveletekit example
Suggestion to using vite instead of svelte-kit. This will be deprecated in the future, and things work quite nicely with vite.

https://github.com/sveltejs/kit/pull/5332

Otherwise folks running the example might run into:

```
svelte-kit dev is no longer available - use vite dev instead

1. Install vite as a devDependency with npm/pnpm/etc
2. Create a vite. config. js with the @sveltejs/kit/vite plugin (see below)
3. Update your package.json scripts to reference 'vite dev' instead of `svelte-kit dev'

// vite. config.js
import { sveltekit } from '@sveltejs/kit/vite';

/** @type fimport('vite').UserConfig} */
  const config = {
  plugins: [sveltekit()]
};

export default config;
```

smaye81:
Fixes #9.

Svelte introduced some breaking changes w/r/t Svelte Kit and Vite.  They now recommend using Vite for various commands and performing a few migration tasks.

This upgrades to latest and fixes the app to be compliant with the new changes.


svelte/src/gen/buf/connect/demo/eliza/v1/eliza_connectweb.ts
```diff
@@ -31,44 +38,43 @@ import {MethodKind} from "@bufbuild/protobuf";
...
-     * response given a one sentence request.
-     *
-     * @generated from rpc buf.connect.demo.eliza.v1.ElizaService.Say
-     */
-    say: {
-      name: "Say",
-      I: SayRequest,
-      O: SayResponse,
-      kind: MethodKind.Unary,
+    typeName: 'buf.connect.demo.eliza.v1.ElizaService',
```
timostamm, line 41: Looks like this file was reformatted.  What do you think about adding some `.prettierignore` rules specific to the generated code to this repository? 

It's good practice because it avoids random diffs and doesn't waste cycles, so we might just as well recommend the practice by using it here.

timostamm (approved):

svelte/src/gen/buf/connect/demo/eliza/v1/eliza_connectweb.ts
smaye81: We actually do have the `**/gen` path ignored in the root `.prettierignore` file.  But for some reason, the `npm run format` and `npm run lint` scripts are not picking it up.  From what I'm reading there's a bit of an issue with how Prettier resolves its ignore files:

https://github.com/prettier/prettier/issues/4081

Also see the [CLI docs](https://prettier.io/docs/en/cli.html#--ignore-path).

I updated the scripts to specify the path to the ignore file.  Not a great solution but probably an acceptable tradeoff here.

mfridman (approved):

svelte/src/gen/buf/connect/demo/eliza/v1/eliza_connectweb.ts
timostamm: Yeah, in a typical monorepo, you would have a single config at the root, and invoke prettier from there. But this is a collection of standalone projects. It's understandable that prettier doesn't look in the parent (it wouldn't be able to determine project boundaries). To keep the examples self-contained, I'd recommend duplicating the config files. 

Definitely not crucial :slightly_smiling_face:

PR #18, opened 2022-11-01
Hardcode all packages under dependabot config

rubensf:
Dependabot doesn't inheretly find all nested package.json, and in this case it's complaining it can't find any [1]. It also doesn't support wildcards, so the hack scripts options to handle that are complex [2].

We could add some generate/lint rules to ensure the dependabot configuration is up to date, but that seems overkill for this repo.

1: https://github.com/bufbuild/connect-web-integration/network/updates/496698197
2: https://github.com/dependabot/dependabot-core/issues/2178#issuecomment-742259742

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,30 +12,6 @@ updates:
       timezone: UTC
       time: "07:00"
   - package-ecosystem: "npm"
-    directory: "svelte"
-    schedule:
-      internal: "monthly"
-      timezone: UTC
-      time: "07:00"
-  - package-ecosystem: "npm"
-    directory: "react-native"
-    schedule:
-      internal: "monthly"
-      timezone: UTC
-      time: "07:00"
-  - package-ecosystem: "npm"
-    directory: "plain"
-    schedule:
-      internal: "monthly"
-      timezone: UTC
-      time: "07:00"
-  - package-ecosystem: "npm"
-    directory: "vue"
-    schedule:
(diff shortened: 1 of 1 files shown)


smaye81:
Actually we may want to just manually update these ourselves.  This repo (as you've discovered) has a ton of nested projects that are all separate, all with their own package.json files.  The Dependabot PRs will probably be extremely onerous to handle.  

Anyway we could maybe turn off Dependabot for this repo (asking from a compliance standpoint).

rubensf:
I think we can get away without dependabot, but I wonder if it wouldn't be useful to keep it for the reminders? There's a couple of our repos where do we exactly that - we don't even want to rely on dependabot to actually do the merges, but it's a useful tracker for "out of date" things.

smaye81:
That works, yeah.  How does it work as a reminder?  Does Dependabot still create a bunch of PRs?

rubensf:
... Yes, I thought of it as creating a bunch of PRs 🥲

smaye81:
Hmm ok.  That seems fine I guess.  The bot runs monthly so maybe it won't be too noisy.  We can try it out.

Super dumb nit but could you put the list in alphabetical order by directory?  It will make it much easier to add / remove things if need be.

rubensf:
Good ask - just made it all in alphabetical order.

If the PRs become a nuisance we can turn it off :)

smaye81 (approved):
Thank you sir.

PR #72, opened 2022-11-01
Angular upgrade

smaye81:
Updates the Angular project to all latest deps.

Tested by running all commands in `package.json` to verify they all still work.


angular/package.json
```diff
@@ -11,33 +11,33 @@
...
+    "@angular/core": "^14.2.8",
+    "@angular/forms": "^14.2.8",
+    "@angular/platform-browser": "^14.2.8",
+    "@angular/platform-browser-dynamic": "^14.2.8",
+    "@angular/router": "^14.2.8",
     "@bufbuild/connect-web": "^0.3.1",
     "@bufbuild/protobuf": "^0.2.1",
     "@bufbuild/protoc-gen-connect-web": "^0.3.1",
     "@bufbuild/protoc-gen-es": "^0.2.1",
-    "rxjs": "~7.5.0",
```
seankimdev, line 26: Are we okay with several version changes from `~version` to `^version` including this one? Just making sure.

smaye81: Yeah this is fine.  This is just an example repo for showing how to integrate connect-web, so these should be fine.

seankimdev (approved):

PR #207, opened 2022-12-29
Add a Webpack CommonJS project

smaye81:
This adds a project which uses Webpack and CommonJS imports so that Connect-Web's CommonJS-structured code can be tested.


react/webpack-cjs/webpack.config.js
```diff
@@ -0,0 +1,38 @@
...
+        use: ["style-loader", "css-loader"],
+      }
+    ]
+    },
+    resolve: {
+        extensions: ['.tsx', '.ts', '.js'],
+        extensionAlias: {
+              '.js': ['.ts', '.js'],
+        },
+    }
```
timostamm, line 37: If I remember correctly, this is needed for the import statements with a `.js` extension that we generate?

timostamm (approved):
If I understand correctly, this works out of the box - with the exception of import path extensions? 

> Connect-Web's CommonJS-structured code

Maybe worth rephrasing a bit. The generated code still uses ESM. But we are testing the CommonJS exports of @bufbuild/protobuf and @bufbuild/connect-web.

react/webpack-cjs/webpack.config.js
smaye81: Yep, that's correct.

PR #208, opened 2022-12-30
Add Makefile and unit tests

smaye81:
This adds a Makefile and unit tests to every project so that we can quickly run tests, builds, and library updates with a single make command.


react/parcel/package.json
```diff
@@ -4,22 +4,27 @@
   "description": "",
   "source": "src/index.html",
   "scripts": {
-    "start": "npm run buf:generate && npx parcel src/index.html -p 3000",
-    "build": "npm run buf:generate && npx parcel build src/index.html",
-    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "unset NODE_OPTIONS && npm run buf:generate && npx parcel src/index.html -p 3000 --log-level error",
```
timostamm, line 7: Why do we `unset NODE_OPTIONS`?

angular/package.json
```diff
@@ -2,12 +2,12 @@
   "name": "buf-ng",
   "version": "0.0.0",
   "scripts": {
-    "start": "ng serve --port 3000",
-    "build": "ng build",
-    "watch": "ng build --watch --configuration development",
-    "test": "ng test",
+    "start": "npx ng serve --port 3000",
```
timostamm, line 5: I'm surprised to see `npx` in an npm script. npm script already inherit from the npm environment, e.g. have "binaries" on the $PATH. Is there a specific reason we're adding this?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,24 @@
...
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.tmp
+            .tmp
+          key: ${{ runner.os }}-connect-web-integration-ci-${{ hashFiles('Makefile') }}
+          restore-keys: |
+            ${{ runner.os }}-connect-web-integration-ci-
+      - name: make
+        run: make
```
timostamm, line 24: Would it be a bit more clear that this runs only tests (and not the updates) if it was `make test`?

timostamm (commented):
This will be incredibly helpful in maintaining this repository 🙏

react/parcel/package.json
smaye81: React Native scripts require it to be set like `export NODE_OPTIONS=--openssl-legacy-provider`, but if it's still set for some of the other projects, the command errors out, so it needs unset.  If there's a better way to do this, I'm all ears.  We could do it in the Makefile, but then it leaves the package.json files vulnerable to this error.

angular/package.json
smaye81: Ah no reason.  I removed the `npx` occurrences from all the package.jsons.

react/parcel/package.json
timostamm: Can we somehow only set NODE_OPTIONS for react-native then? I think users will wonder about `unset NODE_OPTIONS` when they look up this parcel example or others.

smaye81: Yep, I removed the unsetting in the scripts using it.  The issue could still occur if you manually set it and then run parcel or vue, but the overall `make test` run seems to work.

timostamm: I think most users will come here for a specific framework. Seems fine to me.

timostamm (approved):

PR #443, opened 2023-03-09
Updates Next.js example to be full-stack

smaye81:
This updates the Next.js example to be full-stack, utilizing `connect-node` with a Next.js custom server.  Note that this uses a `NextJsApiAdapter` implementation to make it all flow together.  This adapter will eventually be offered as a package from `connect-es`, but can maybe start here as a testing ground.


nextjs/tsconfig.json
```diff
@@ -13,8 +17,20 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "incremental": true
+    "incremental": true,
+    "plugins": [
+      {
+        "name": "next"
```
timostamm, line 23: Why is the plugin added here? It doesn't look like it's part of a default create-next-app?

nextjs/server.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const fallback = app.getRequestHandler()
+const apiRoute = nextJsApiRouter({
+    routes,
+    fallback,
+})
+
+app.prepare().then(() => {
+    // @ts-ignore
+    createServer(apiRoute.handler)
```
timostamm, line 20: A custom server is an advanced feature, so it's not going to be widely used, which makes this very difficult for users to follow. Custom servers can't be deployed to Vercel. Why do we use one here?

timostamm (changes requested):
I ran `npm ci` and `npm start`, but don't get a response on localhost:3000, and there are a couple of other small issues. 

But I also think that there is a larger issue with this approach: This project deviates a _lot_ from default create-next-app output. It uses the experimental `app/` directory, a custom server, and it changes default npm scripts. 

How should users adapt this to their existing project? How do they find out how to setup a new project this way? I don't think this is a good approach for an example. 

Let's keep this as simple as possible please, and stick to the defaults of create-react-app (in a current version). Literally just running `npx create-next-app nextjs` and hitting <kbd>Enter</kbd> until it's done should be the starting point, and any necessary changes need to be documented, so that we have a good understanding of friction points.

nextjs/tsconfig.json
smaye81: These values are added by Next.js:

```
We detected TypeScript in your project and reconfigured your tsconfig.json file for you. Strict-mode is set to false by default.

The following suggested values were added to your tsconfig.json. These values can be changed to fit your project's needs:

        - include was set to ['next-env.d.ts', '.next/types/**/*.ts', '**/*.ts', '**/*.tsx']
        - plugins was updated to add { name: 'next' }
```

nextjs/server.ts
smaye81: Users have asked for a full-stack version of Next.js and I assumed a custom server was a way to show how that could be done.

PR #444, opened 2023-03-10
feat: Add fastify-cors example

akinorimizushima:
This will add an example using fastify-cors which can receive requests from different hosts or ports.


node.js/fastify-cors/server.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const server = fastify()
+
+await server.register(cors, {
+    origin: "*",
+    allowedHeaders: [
+        "content-type",
+        "Connect-Protocol-Version",
+        "Access-Control-Allow-Origin",
+    ],
```
timostamm, line 29: We need a couple more headers here, and we also need to set `exposedHeaders` and `methods`. 

But with https://github.com/bufbuild/connect-es/pull/526, you can simply use the values we export.

timostamm (commented):
Thank you for the contribution, @akinorimizushima!

We have been discussing how we can make setting up CORS a bit easier. The PR https://github.com/bufbuild/connect-es/pull/526 adds a nice little helper for that. Do you mind waiting for that PR to land so you can use it here?

akinorimizushima:
Hi @timostamm, thanks for heads up. Okay, will work on it once the feature is published.

akinorimizushima:
@timostamm Pushed some commits using the upgraded version but https://github.com/bufbuild/connect-es-integration/pull/463 is doing the similar thing. So feel free to reject it if this PR is not helpful as an example.

smaye81:
> @timostamm Pushed some commits using the upgraded version but #463 is doing the similar thing. So feel free to reject it if this PR is not helpful as an example.

Hey @akinorimizushima. This is definitely a helpful example. What we are trying to do is make each example (express, fastify, etc.) self-sufficient and showing both a full-stack as well as a CORS example. #463 is an attempt to unify the Express project so that we don't have `express`, `express-cors`, `fastify`, `fastify-cors`, etc. Once we land #463, we're hoping to have a good pattern to follow and then we'd appreciate it if you could modify this to follow that same pattern with fastify.

smaye81:
Hi @akinorimizushima. On second thought, we decided to combine these changes into #463 since that PR was already doing a massive refactoring of the Node examples. Your work on `fastify` was ported to that PR, so I am going to close this in favor of #463. Thank you for the contribution!

PR #463, opened 2023-03-15
Sayers/express cors

smaye81:


node.js/README.md
```diff
@@ -4,8 +4,8 @@ This project shows examples of Connect for Node.js working in concert with Conne
```
smaye81, line 4: Will update the READMEs.  More interested in whether this setup looks good.

.prettierrc
```diff
@@ -1,6 +1,5 @@
 {
     "trailingComma": "es5",
     "tabWidth": 4,
-    "semi": false,
```
timostamm, line 4: Is there a reason we use non-default prettier setup? 

Quoting from [their docs](https://prettier.io/docs/en/option-philosophy.html):

> By far the biggest reason for adopting Prettier is to stop all the ongoing debates over styles.

vanilla-node/server.ts
```diff
@@ -50,19 +54,30 @@ const handler = connectNodeAdapter({
...
+    exposedHeaders: [...connectCors.exposedHeaders],
+});
 
-http.createServer(handler).listen(3000, () => {
-    stdout.write('The server is listening on http://localhost:3000\n')
-    stdout.write('Run `npm run client` for a terminal client.\n')
-})
+http.createServer((req, res) => {
+    corsHandler(req, res, () => handler(req, res));
+}).listen(PORT, () => {
```
timostamm, line 79: Can you add an example for starting a server _without_ CORS? I think it should fit on a single line, so it could be in a comment, like "If you don't need the CORS middleware, simply use ...".

fastify/server.ts
```diff
@@ -0,0 +1,61 @@
...
+import { fastifyConnectPlugin } from '@bufbuild/connect-fastify';
+import { cors as connectCors } from '@bufbuild/connect';
+import fastifyCors from '@fastify/cors';
+import { readFileSync } from 'fs';
+import * as esbuild from 'esbuild';
+import routes from './connect';
+
+const server = fastify();
+
+await server.register(fastifyCors, {
```
timostamm, line 25: Can you add a short comment explaining what this is?

express/server.ts
```diff
@@ -0,0 +1,78 @@
...
+import { cors as connectCors } from '@bufbuild/connect';
+import { expressConnectMiddleware } from '@bufbuild/connect-express';
+import { readFileSync } from 'fs';
+import { stdout } from 'process';
+import * as esbuild from 'esbuild';
+import routes from './connect';
+
+const PORT = 3000;
+
+const corsOptions: cors.CorsOptions = {
```
timostamm, line 27: Can you add a short comment explaining what this is?

timostamm (approved):

PR #465, opened 2023-03-15
Add integration with @bufbuild/connect-next

smaye81:
This adds an integration with the new [@bufbuild/connect-next](https://www.npmjs.com/package/@bufbuild/connect-next) package to show a full-stack example with Next.js.


nextjs/pages/index.tsx
```diff
@@ -28,7 +28,7 @@ function App() {
     const client = createPromiseClient(
         ElizaService,
         createConnectTransport({
-            baseUrl: 'https://demo.connect.build',
+            baseUrl: 'http://localhost:3000/api',
```
timostamm, line 31: To make this work when publishing on vercel without modifying the source just to deploy:

```suggestion
            baseUrl: '/api',
```

timostamm (approved):

PR #605, opened 2023-05-19
Make a sveltekit-ssr example (universal and server-only)

dimitropoulos:
This PR implements a sveltekit example demonstrating how to do SSR in both styles that sveltekit supports (both of [Universal and Server-only](https://kit.svelte.dev/docs/load#universal-vs-server)).


svelte/sveltekit-ssr/src/routes/server-ssr/+page.svelte
```diff
@@ -0,0 +1,181 @@
+<script lang="ts">
+    import { ElizaService } from '../../gen/buf/connect/demo/eliza/v1/eliza_connect.js';
+    import { IntroduceRequest } from '../../gen/buf/connect/demo/eliza/v1/eliza_pb.js'
+    import { createClient, wrapFetch } from '../../utils.js';
+
+    export let data: { serverSentence: string } = {
```
paul-sachs, line 6: I think we can actually infer this by doing:

```ts
import type { PageData } from "./$types";
export let data: PageData = { ... }
```

This leverages the generated types.

svelte/sveltekit-ssr/src/routes/universal-ssr/+page.svelte
```diff
@@ -0,0 +1,181 @@
+<script lang="ts">
+    import { ElizaService } from '../../gen/buf/connect/demo/eliza/v1/eliza_connect.js';
+    import { IntroduceRequest } from '../../gen/buf/connect/demo/eliza/v1/eliza_pb.js'
+    import { createClient, wrapFetch } from '../../utils.js';
+
+    export let data: { universalSentence: string } = {
```
paul-sachs, line 6: Ditto as above.

paul-sachs (approved):
This is a great example! I wish we had a better CRUD service to better show server side rendering but it works!!!

svelte/sveltekit-ssr/src/routes/server-ssr/+page.svelte
dimitropoulos: hmm.  I don't see that: all I see is

```
(alias) type PageData = {
  [x: string]: any;
}
```

paul-sachs: Weeeeeird, I def see:

image

paul-sachs: I wonder if that's due to the sveltejs for vscode extension... but that would be weird.

timostamm (commented):
Nice, but I don't think the universal `load` function is working. The log statements show up in the browser and in the console, so the function is executed on the server and on the frontend, which is expected. 

But it is issuing a HTTP request (with `npm start` but also with `npm run build && npm run preview`):

Screenshot 2023-05-22 at 09 52 09

I would expect svelte to record the response on the server side, and embed it in the web page. When the document loads in the browser, I would expect svelte to run the `load` function and hydrate the fetch response from the embedded data instead of issuing a HTTP request.

If this does not work, it would be nice to understand why, and whether there is a reasonable way to work around it for the user or for us. We should not give an example that does not work as expected.

For the server load function, we rely on page data to be serialized and parsed again. What are the constraints with protobuf here? I imagine that `bigint` and classes might not be well supported.

svelte/sveltekit-ssr/src/routes/server-ssr/+page.svelte
dimitropoulos: I have no idea why, but I opened it up this morning and it's working now.  welp.  gamma rays. 40b9417cc3a3ef88702d740ff58928e4598429cd

svelte/sveltekit-ssr/src/routes/universal-ssr/+page.svelte
dimitropoulos: fixed in d38979d280b1c31e1a51f5c73add7980b6d7ca07

dimitropoulos:
> from the embedded data instead of issuing a HTTP request

what gives you this expectation?

What I see in the [docs](https://learn.svelte.dev/tutorial/universal-load-functions) are
> Now, the functions will run on the server during server-side rendering, but will also run in the browser when the app hydrates or the user performs a client-side navigation.

[and](https://kit.svelte.dev/docs/load#universal-vs-server-when-does-which-load-function-run)
> By default, universal load functions run on the server during SSR when the user first visits your page. They will then run again during hydration, reusing any responses from [fetch requests](https://kit.svelte.dev/docs/load#making-fetch-requests).

which

timostamm:
With a plain fetch, using Response.json(), I see hydration data in the page source, and no additional request from the client.

<details>
<summary>+page.ts</summary>

```ts
import { SayRequest } from '../../gen/buf/connect/demo/eliza/v1/eliza_pb'
import type { PageLoad } from './$types'
export const load: PageLoad = async ({ fetch }) => {
  const r = await fetch("https://demo.connect.build/buf.connect.demo.eliza.v1.ElizaService/Say", {
    method: "POST",
    headers: {
      "Content-Type": "application/json"
    },
    body: new SayRequest({
      sentence: "hi from the server",
    }).toJsonString()
  });
  const body = await r.json();
  console.log("server sentence:", body.sentence)
  return { universalSentence: body.sentence }
}
```
</details>

<details>
<summary>Network inspector w/o additional requests</summary>

Screenshot 2023-05-22 at 18 08 50
</details>



<details>
<summary>Hydration data in the document</summary>

```html
	<script type="application/json" data-sveltekit-fetched data-url="https://demo.connect.build/buf.connect.demo.eliza.v1.ElizaService/Say" data-hash="1y6nxzi">{"status":200,"statusText":"OK","headers":{"content-type":"application/json"},"body":"{\"sentence\":\"Hi there...how are you today?\"}"}</script></div>
    </body>
</html>
```
</details>

timostamm:
The SSR example in https://github.com/bufbuild/connect-es-integration/pull/610 turned out nice and simple. I think we should take the same approach here, specifically:
1. add the server-only SSR example as a separate page inside the existing project instead of adding the new one, and do only SSR on the page
2. show passing a protobuf message and passing individual properties in the server-only example, like we did for getServerSideProps
3. add the universal-load example a separate page in the existing project too, but pass the protobuf message (class instance) through without the toJson workaround, because that should work with universal-load
4. add a navigation to switch between the pages

Regarding the example for universal load, I still do not believe that it actually works. I think we might be able to make a change in @bufbuild/connect-web specifically to support it, and thereby also helping out with https://github.com/bufbuild/connect-es-integration/pull/602 and https://github.com/bufbuild/connect-es/issues/199. I'm cool with merging it anyhow so that we have a reproducible test case, it just needs a TODO explaining that it isn't currently supported and that we are looking into it. If it turns out that is not feasible, we can still remove the example again.

PR #608, opened 2023-05-25
[experiment] uses superjson to break SSR boundary

dimitropoulos:
(this is just a proof of concept PR intended to just explore what's possible with SSR)

this is a stacked PR off of #607, so for the important changes see https://github.com/bufbuild/connect-es-integration/pull/608/commits/72b37bcfb701fd8e138cd6d8086b1d7b23443ac3

`BigInt`, `UInt8Array` (protobuf's `bytes` type), and our `Message` methods won't pass the SSR boundary.  We've tried (and failed) now with Svelte as well as Next.JS.  Both assume you're passing regular natively javascript-compatible JSON over the wire (i.e. numbers are not BigInts, etc.).

I wanted to see how far I could push that SSR boundary (preferably, all the way) and it looks like I succeeded.  Observe this example (in `nextjs/ssr`) which demonstrates "reviving" the objects after passing them through the SSR boundary.

It's not great to have to include `superjson`, but this is at least the first thing we've tried that allows SSR to work 100% as normal with all the same abilities as protobuf itself.

----

## Next things to try

1. https://www.npmjs.com/package/next-superjson
2. making something exactly like `next-superjson` that plugs into `Next.JS` but just goes through our entire stack (thereby potentially allowing us to drop `superjson` from the equation entirely).


nextjs/ssr/pages/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+    responseRepeatedSfixed32,
+    responseRepeatedSfixed64,
+    responseRepeatedBool,
+    responseRepeatedString,
+    responseRepeatedBytes,
+  };
+  
+  return {
+    props: {
+      superjsonResult: SuperJSON.serialize(kitchenSinkResponse),
```
timostamm, line 120: Couldn't you simply convert to JSON here...

```suggestion
      superjsonResult: kitchenSinkResponse.toJson(),
```

nextjs/ssr/pages/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+};
+
+const revive = <T extends new (...args: any[]) => any>(ResponseClass: T, superjsonResult: SuperJSONResult) => {
+  return new ResponseClass(SuperJSON.deserialize<T>(superjsonResult))
+}
+
+export default function Page({
+  superjsonResult,
+}: InferGetServerSidePropsType<typeof getServerSideProps>) {
+  const ssrOutput = revive(KitchenSinkResponse, superjsonResult);
```
timostamm, line 132: And parse again with:

```suggestion
  const ssrOutput = KitchenSinkResponse.fromJson(superjsonResult);
```

?

nextjs/ssr/pages/index.tsx
dimitropoulos: yes! see 1bbd57733081b8b17162c47a4767ba4b4469a04d

nextjs/ssr/pages/index.tsx
dimitropoulos: ahh yes, thanks! see 1bbd57733081b8b17162c47a4767ba4b4469a04d

dimitropoulos:
1bbd57733081b8b17162c47a4767ba4b4469a04d is a great usage example of doing the same concept, but using toJson and fromJson


Screenshot 2023-05-30 at 13 04 43

PR #610, opened 2023-05-30
Add example for SSR with Next.js

dimitropoulos:
This PR adds a small navigation to the existing Next.js example.

Upon clicking that link, you are shown a small SSR example page:

> note: [previous attempt](https://github.com/bufbuild/connect-es-integration/pull/607)


nextjs/pages/ssr.tsx
```diff
@@ -0,0 +1,41 @@
...
+    // note: you cannot use a relative path like `/api` here because SSR will break.  SSR requires absolute URLs.
+    baseUrl: "https://demo.connect.build",
+    fetch: wrapFetch("calling from getServerSideProps", fetch),
+  });
+  const client = createPromiseClient(ElizaService, transport);
+  const request = { sentence: "hi (from the server)" };
+  const response = await client.say(request);
+
+  return { props: { request, response: { sentence: response.sentence } } };
+  //                         ^ note: you can use `response.toJson()` instead, but you will lose all type information.
```
timostamm, line 19: Looks to me like making a call is straight-forward, but passing the results to the page is the challenging part. Similar to the Svelte server load case in https://github.com/bufbuild/connect-es-integration/pull/605.

---

We just talked offline and will meet to discuss the best approach.

nextjs/pages/ssr.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { createConnectTransport } from "@bufbuild/connect-web";
+import styles from '../styles/Eliza.module.css';
+import { wrapFetch } from "../utils";
+
+export const getServerSideProps = async () => {
+  const transport = createConnectTransport({
+    // note: you cannot use a relative path like `/api` here because SSR will break.  SSR requires absolute URLs.
+    baseUrl: "https://demo.connect.build",
+    fetch: wrapFetch("calling from getServerSideProps", fetch),
+  });
```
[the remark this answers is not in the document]
timostamm: Gotcha. Best to remove it. I think it's clear that getServerSideProps runs on the server.

nextjs/pages/ssr.tsx
timostamm: As we just discussed offline, in some cases, passing individual properties works out, but in others, it would be terribly verbose. 

We should give users enough information to make the choice themselves, but definitely show a full example for to toJson() / fromJson() approach so that it's obvious how to use it.

dimitropoulos: sure thing: 900a8111fb2ae61d05b2d065b212139b225b3361

timostamm (approved):
LGTM!

PR #687, opened 2023-06-22
Simplify nextjs example

paul-sachs:
- Splitting example into streaming and unary example
- Simplify message sending via form submit


nextjs/pages/index.tsx
```diff
@@ -1,118 +1,90 @@
...
-                <Link href="/ssr">View SSR Example</Link>
+                <h4 className={styles.subtitle}>
+                    Next.js + Client-side Fetching
+                </h4>
+                <div className={styles.links}>
+                    <Link href="/ssr">View SSR Example</Link>
+                    <Link href="/server-streaming">
+                        View Server Streaming Example
+                    </Link>
+                </div>
```
timostamm, line 56: Can we change the header to provide a better navigation?

Screenshot 2023-06-23 at 17 54 50

timostamm (commented):
Looks great to me, thank you!

nextjs/pages/index.tsx
paul-sachs: Yeah i'll make those changes, makes sense.

paul-sachs: Sorry for the delay. Updated!

smaye81:
I was going to say let's change to the `connectrpc` paths here, but let's just merge this and then we can update them all in one PR (#777).

smaye81 (approved):

PR #777, opened 2023-07-28
Migrate repo to use connectrpc paths

smaye81:
This migrates the repo and all examples to use the new `connectrpc` paths instead of `bufbuild` and `demo.connect.build`


timostamm:
@paul-sachs, can you wrap up https://github.com/bufbuild/connect-es-integration/pull/687 so that we don't have to update once?

timostamm (approved):

PR #911, opened 2023-08-24
Add an example using RTL with mocked transports to react/cra

smaye81:
This adds an example of component testing using React Testing Library to the Create React App project.

In doing so, it refactors the components in the project so that a `Transport` interface is passed as a prop. This way, users can easily use `createRouterTransport` to create a mock transport and pass that to the component in their test. This approach can then be suggested in docs as a recommendation for easily testing components that interact with Connect clients and transports.

Note that this also moves the actual test execution out of `react-app-rewired` and into pure Jest. This is because we need to set the `testEnvironment` to `@bufbuild/jest-environment-jsdom` because of the issues with `jsdom`. However, neither CRA nor `react-app-rewired` seem to allow you to override this value despite docs indicating it is possible. It is also just more intuitive and easy to reason about to use Jest directly.


smaye81:
@timostamm Would also like to add a testing example (to one of the other React projects -- esbuild maybe) to show an example of testing with Jest snapshots. Are there other examples you'd like to see?

react/cra/package.json
```diff
@@ -48,12 +41,16 @@
     "web-vitals": "^3.4.0"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
```
timostamm, line 44: Is this necessary?

react/cra/src/App.tsx
```diff
@@ -1,112 +1,120 @@
...
-    sender: 'eliza' | 'user'
+    text: string;
+    sender: "eliza" | "user";
 }
 
-function App() {
-    const [statement, setStatement] = useState<string>('')
-    const [introFinished, setIntroFinished] = useState<boolean>(false)
+interface ChatContainerProps {
+    transport: Transport;
```
timostamm, line 14: This certainly works, but in a larger application, it leads to "prop drilling", which is considered an anti-pattern in react.

Can we adapt the code snipped from the docs section [Managing clients and transports](https://connectrpc.com/docs/web/using-clients#managing-clients-and-transports) to support a Transport that is overridden in tests?

Ideally, we would be able to add such a code snippet in [Testing / Mocking services](https://connectrpc.com/docs/web/testing#mocking-services) and link to this example.

timostamm (commented):
> Would also like to add a testing example (to one of the other React projects -- esbuild maybe) to show an example of testing with Jest snapshots. 

Sounds like a good idea to me if it is easy to set up. But more importantly, we need an idiomatic example for overriding the transport in tests, see comment below.

react/cra/package.json
smaye81: Without it, I was seeing the following when building:

```
One of your dependencies, babel-preset-react-app, is importing the
"@babel/plugin-proposal-private-property-in-object" package without
declaring it in its dependencies. This is currently working because
"@babel/plugin-proposal-private-property-in-object" is already in your
node_modules folder for unrelated reasons, but it may break at any time.

babel-preset-react-app is part of the create-react-app project, which
is not maintianed anymore. It is thus unlikely that this bug will
ever be fixed. Add "@babel/plugin-proposal-private-property-in-object" to
your devDependencies to work around this error. This will make this message
go away.
```

react/cra/config-overrides.js
```diff
@@ -1,10 +1,11 @@
...
 
 // config-overrides.js
-module.exports = function override(config, env) {
-    config.resolve = {
-        plugins: [new ResolveTypeScriptPlugin()],
-    }
-
-    return config
-}
+module.exports = {
```
smaye81, line 4: This is unrelated to this PR, but I think it's much clearer what we're overriding here if we use this syntax (an object with `webpack` as a property).

timostamm (approved):
The use of React looks appropriate here. This also means that this example should translate well to @connectrpc/connect-query, which already uses a context for the transport.

I have the feeling that the API would be more convenient if createRouterTransport() took an approach more similar to the [Playwright integration](https://github.com/connectrpc/connect-playwright-es#usage), but we can easily follow up with an alternative function later down the road.

PR #912, opened 2023-08-28
Add component testing example using Vue and mocked transports

smaye81:
This modifies the Vue example to add component testing using a mocked transport. To do so:

* Uses Vue's provide/inject approach to provide the transport to the component
* Mocks the provider in unit tests
* Uses Vue Test Utils to verify the DOM


vue/src/views/__tests__/ElizaView.spec.ts
```diff
@@ -1,11 +1,73 @@
...
+        let fi = wrapper.findAll(".eliza-resp-container p");
+
+        expect(fi.at(0)?.text()).toContain("What is your name?");
+
+        // Enter a name in the input and click send
+        await input.setValue("Steve");
+        await sendButton.trigger("click");
+
+        // Wait for the introduce response from our mocked Eliza and verify it is our mocked response
+        await flushPromises();
```
smaye81, line 58: https://test-utils.vuejs.org/guide/advanced/async-suspense.html#resolving-other-asynchronous-behavior

vue/src/views/__tests__/ElizaView.spec.ts
```diff
@@ -1,11 +1,72 @@
...
+describe("ElizaView", () => {
+    let input: DOMWrapper<Element>;
+    let sendButton: DOMWrapper<Element>;
+    let wrapper: VueWrapper<any, any>;
+
+    beforeEach(() => {
+        wrapper = mount(ElizaView, {
+            global: {
+                provide: {
+                    [keys.TRANSPORT as symbol]: mockTransport,
```
timostamm, line 38: Seems a bit awkward. I'm not sure why InjectionKey is extending Symbol instead of an intersection type. 

Using 
```ts
export const transportKey: InjectionKey<Transport> & symbol = Symbol();
```

and then

```ts
[transportKey]: mockTransport,
```

here seems less gross, though.

smaye81: I would say both look gross, tbh. Plus, if we're trying to stay close to the docs as with the `keys` export, then I think I'd prefer going with what I have since that follows the docs.

vue/package.json
```diff
@@ -2,13 +2,13 @@
   "name": "buf-vue",
   "version": "0.0.0",
   "scripts": {
-    "start": "vite",
+    "start": "vite --port 3000",
```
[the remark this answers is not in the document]
smaye81: All examples in the repo use port 3000 for the web client. Most of them default to that, but we manually set this in a few other places (Angular, Vite, Svelte, the Node examples). I think this small deviation is probably a fair tradeoff for consistency with all the other examples. Plus, then we'd have to go and change the other scripts that manually set them and I don't think that's worth the effort.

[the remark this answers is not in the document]
timostamm: They should all use the standard behavior. We can clean that up separately.

vue/src/views/__tests__/ElizaView.spec.ts
smaye81: For the grossness, I meant the export declaration. `[transportKey]` is definitely better (even though this would probably be in a `beforeEach` and not duplicated that much anyway). Regardless, here's the docs for typing that export: 

https://vuejs.org/guide/typescript/composition-api.html#typing-provide-inject

timostamm: When writing tests, you typically want to test different situations in the backend, and you will need to create and inject the mock transport for each situation. That's why I think it's best to keep the call site as simple as possible, at the cost of complexity at the definition of the key.

The Vue.js docs are lacking because the typing for InjectionKey is broken. See [here](https://github.com/microsoft/TypeScript/issues/54885#issuecomment-1624286684) for an explanation from the TS lead. 

You could add a comment on the key declaration linking to https://github.com/vuejs/core/issues/8744.

timostamm (approved):

PR #914, opened 2023-08-29
Add component testing example to Svelte

smaye81:
This adds a component testing example to the Svelte app using `@testing-library/svelte` and some suggested paradigms in Svelte documentation. This involved:

* Previous `npm run test` script was using Playwright to run e2e tests, so that was moved to an `e2e/` directory and invoked from a `test:e2e` script so that unit tests could be run via `npm run test`.
* The transport for the Connect client is now set/accessed from Svelte's Context API. Since the Context API can only be invoked within a component lifecycle, the calls to `setContext` were moved to a Svelte layout as illustrated [here](https://kit.svelte.dev/docs/state-management#using-stores-with-context).
* The actual test, in turn, then modifies this value in context through the use of a parent helper component which allows for passing in a component under test and a mock context to be used during the test run. This example was pulled from the Svelte Society recipes [here](https://github.com/svelte-society/recipes-mvp/blob/master/testing.md#testing-the-context-api). The Svelte Society page is not officially affiliated with the Svelte maintainers, but is linked from their site as a third-party resource.
* Svelte's component lifecycle within a test seems to be tripping over either completely flushing promises or calling a generator function (i.e. Introduce). The library [flush-promises](https://www.npmjs.com/package/flush-promises), which is recommended for async Vue testing, was introduced to force these promises to resolve.

In addition, the `npm run lint` command was broken due to some outdated Svelte plugins, so the latest config was migrated for ESLint and all commands were verified to be working.


timostamm (commented):
Looks great, just one optional comment from me. Thanks for the comprehensive PR description.

`npm run lint` is still broken though after a `npm run build`. Every project needs it's own .prettierignore file if uses prettier and needs exclusions.

smaye81:
> Looks great, just one optional comment from me. Thanks for the comprehensive PR description.
> 
> `npm run lint` is still broken though after a `npm run build`. Every project needs it's own .prettierignore file if uses prettier and needs exclusions.

Yep, good call. Fixed.

timostamm (approved):

PR #983, opened 2023-09-04
Add examples for testing clients and servers on Node.js

timostamm:
This adds exemplary tests for a service and a client application.


srikrsna-buf (approved):
The mocking bits are not easy to parse but I don't see an alternative.

timostamm:
> The mocking bits are not easy to parse but I don't see an alternative.

Can you elaborate which ones specifically?

vanilla-node/client.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    it("should ask for username and send it to the server", async (t) => {
+        const io = {
+            question: t.mock.fn<readline.Interface["question"]>(undefined, () => Promise.resolve("Joe")),
+            write: t.mock.fn<readline.Interface["write"]>()
+        };
+        const introduce = t.mock.fn<MethodImpl<typeof ElizaService.methods.introduce>>(
+            async function* introduce() {
+                //
+            }
+        );
```
srikrsna-buf, line 55: These ones, I mean just the types

timostamm: Agreed. I went through a couple of alternatives, but this felt like the least bad option. I think if we didn't use Eliza here (we have been playing around with a stateful "user service" before), the mock might also be easier to follow.

I think long-term, we might also want to invert control with the router transport, and have an API more similar to the [Playwright integration](https://github.com/connectrpc/connect-playwright-es/tree/main#usage). 

Both don't seem easily achievable right now, but IMO definitely both worthwhile to keep in mind.

PR #989, opened 2023-09-05
Add example for testing services with Express

timostamm:


vanilla-node/server.ts
```diff
@@ -17,71 +17,74 @@ import { cors as connectCors } from "@connectrpc/connect";
...
-            res.end();
-          });
-        break;
-      default:
-        res.writeHead(404);
-        res.end();
-    }
-  },
-});
+export function build() {
```
timostamm, line 30: Added this indirection for Fastify, loosely following their [testing guide](https://fastify.dev/docs/latest/Guides/Testing). 

For consistency, I'm doing the same with Express, and updating the Node.js example here to use it as well.

The added tests for "GET /" and "GET /webclient.js" will help us to guard against regressions.

PR #990, opened 2023-09-05
Add CI scripts

### LINKED ISSUE #915: Improve coverage in CI
This repository hosts examples for many different frameworks. Currently, CI is running the following command for an example if any related file was changed:

https://github.com/connectrpc/examples-es/blob/5a9f041cf001e751d2ac1ec706f5e0c5ffec55ea/.github/workflows/ci.yaml#L60

Which runs the npm scripts `buf:generate`, `build`, and `test` after installing dependencies:

https://github.com/connectrpc/examples-es/blob/5a9f041cf001e751d2ac1ec706f5e0c5ffec55ea/manage.mjs#L65-L68

This certainly gives us decent coverage, but looking at the example for Vue.js, we're missing some important scripts:

https://github.com/connectrpc/examples-es/blob/5a9f041cf001e751d2ac1ec706f5e0c5ffec55ea/vue/package.json#L11-L14

E2E tests and the linter never run in CI. Same with the Svelte example, where we also don't run the `check` script, which is important in Svelte because the `build` script does not type-check.

Since the examples do not follow a common pattern with their npm scripts (and they should not), I propose that we add a `ci` script to every example, and run it after installing dependencies. For the Vue.js example, that would be: 

```json
"ci": "npm run buf:generate && npm run lint && npm run build && npm run test && npm run test:e2e:ci"
```

The `lint` script for Vue.js runs eslint with `--fix`. Other examples have a `format` script. To make sure that all files are up to date, I suggest that we add a diff check via git after running the `ci` script ([reference](https://github.com/connectrpc/connect-playwright-es/blob/f3ad0a9dd57f23b11e0e29c4248e8c198cc18b4b/.github/workflows/ci.yaml#L32-L35)).

smaye81:
Fixes #915

This adds a `ci` script to each project and then invokes this script as part of our CI runs.

The `ci` script incorporates all tasks to verify the validity of the project such as `generate`, `test`, `build`, `type-check`, `lint`, and `test:e2e`.

Since each project contains somewhat different scripts, the `ci` script for each project incorporates whichever of the above is defined by the project.


smaye81:
@timostamm running a `checkdiff` is a bit more complicated since we have a monorepo using multiple package managers. The `yarn` projects are all failing because they are not caching the `.yarn/cache` directory and therefore keep installing Linux binaries for `buf` and `esbuild`, which then fails the `checkdiff` check. We can't make use of `setup-node`'s cache because it only allows one package manager to be specified.

We could maybe break up our current `ci.yaml` into `ci-npm.yaml`, `ci-yarn.yaml`, etc. so that we can cache these independently.

smaye81:
They do suggest it if you want to use the zero-installs feature, but in reading more about it, they have a [caveat](https://yarnpkg.com/features/caching#zero-installs) about Github Actions listed there. Plus, it seems to apply to PnP only, so our Unplugged package would be out of luck (unless we committed `node_modules`, which we don't want to do).

I'm just going to add this to .gitignore for both projects since it's probably more trouble than it's worth for our current requirements.

smaye81:
@timostamm should be good now finally.

timostamm (commented):
Let's use yarn version 3.6.1 consistently - in also in package.json and .yarnrc.yaml.

We should be a bit more prudent when adding gitignores for checked in files:
- react/yarn-pnp/.yarn/releases/yarn-3.6.1.cjs
- react/yarn-unplugged/.yarn/releases/yarn-3.2.1.cjs
- react/yarn-unplugged/.yarn/releases/yarn-4.0.0-rc.39.cjs

Please try using corepack instead. See here for reference: https://github.com/connectrpc/connect-query-es/pull/60

smaye81:
I went with `yarn v3.63` since `yarn set version berry` updates to that version. The use of `corepack` handles removing the extraneous release files in `.yarn/releases`.

I also added a version to Remix (`pnpm`).

react/yarn-unplugged/.yarnrc.yml
```diff
@@ -1,2 +1,2 @@
-yarnPath: .yarn/releases/yarn-3.2.1.cjs
 nodeLinker: node-modules
+yarnPath: .yarn/releases/yarn-3.6.3.cjs
```
timostamm, line 2: This is highly confusing when using corepack, because corepack manages package manager versions. Take a look here how it works: https://nodejs.org/api/corepack.html

This line should be removed.

react/yarn-pnp/.yarnrc.yml
```diff
@@ -1 +1 @@
-yarnPath: .yarn/releases/yarn-3.6.1.cjs
+yarnPath: .yarn/releases/yarn-3.6.3.cjs
```
timostamm, line 1: The file can be removed.

manage.mjs
```diff
@@ -184,6 +195,12 @@ class PackageEnt {
         }
     }
 
+    /**
+     */
+    enableCorepack() {
+      this.run(`corepack enable`);
```
timostamm, line 201: Please take a look at how corepack works. It's a one-time setup. 

Is there a specific reason we're not using the approach from https://github.com/connectrpc/connect-query-es/pull/60?

timostamm (commented):
These two files are still checked in but ignored. They should be removed:

- react/yarn-pnp/.yarn/releases/yarn-3.6.3.cjs
- react/yarn-unplugged/.yarn/releases/yarn-3.6.3.cjs

react/yarn-unplugged/.yarnrc.yml
smaye81: Yarn's docs recommend Corepack, but they also recommend using `yarn set version <version>` to update Yarn for a project. So, we can remove this line and the entire file in `yarn-pnp`, but anytime we run `yarn set version <version>` for either of these projects, Yarn is just going to recreate the file again and add that line back.

Similarly re: the `releases` directory, Yarn's [docs](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored) also recommend committing this directory. There's no mention of whether this is advisable when using Corepack, but I don't see an argument against it.

manage.mjs
smaye81: Connect-Query is just one project using pnpm. If we run Corepack one-time for this entire repo, is it going to recursively dive down into each project and enable the package manager correctly? I assumed that since we treat each directory as a separate project, this should be run inside each one.

timostamm: Yes.

From the docs:

> The Corepack proxies will find the closest [package.json](https://nodejs.org/api/packages.html#nodejs-packagejson-field-definitions) file in your current directory hierarchy to extract its ["packageManager"](https://nodejs.org/api/packages.html#packagemanager) property.
> 
> If the value corresponds to a [supported package manager](https://nodejs.org/api/corepack.html#supported-package-managers), Corepack will make sure that all calls to the relevant binaries are run against the requested version, downloading it on demand if needed, and aborting if it cannot be successfully retrieved.

timostamm:
@smaye81, did corepack work out?

smaye81:
It seemed to. I switched our `ci.yaml` to use the same approach connect-query is using and removed it from the `manage.mjs` scripts and all the CI tests are passing. 🤷

react/yarn-pnp/README.md
```diff
@@ -8,6 +8,7 @@ Other tools used:
 
 ## Getting Started
 
+### `corepack enable`
```
timostamm, line 11: Users coming to this yarn example will have yarn installed. Let's remove this instruction, and add the section about corepack from https://github.com/connectrpc/connect-query-es/pull/60/files at the bottom.

Please remove the instructions `yarn set version canary` and `yarn set version berry`.

react/yarn-unplugged/README.md
```diff
@@ -11,6 +11,7 @@ Other tools used:
 
 ## Getting Started
 
+### `corepack enable`
```
timostamm, line 14: See above.

smaye81:
@timostamm updated. Also added the same instructions to the `remix` project.

PR #997, opened 2023-09-15
Update connect packages to `v1.0.0-rc1`

srikrsna-buf:
Update connect packages to `v1.0.0-rc1`


timostamm (commented):
Nice. Is this the result from running `npx @connectrpc/connect-migrate@latest`, as we recommend in the latest release?

srikrsna-buf:
Is that applicable here? I manually updated these. I tried `npx @connectrpc/connect-migrate@latest` now but it says everything is up to date. My understanding is that it only migrates from `@bufbuild/*` packages

timostamm:
Ah, you're right - it upgrades to the RC, but only from @bufbuild 👍

PR #1002, opened 2023-09-29
Don't merge: Demonstrate dual package hazard

timostamm:
This example demonstrates the dual package hazard and verifies the fix by https://github.com/connectrpc/connect-es/pull/842.

For a repeatable example for the dual package hazard with various consumers [see here](https://github.com/connectrpc/examples-es/tree/tstamm/dual-package-hazard-example/dual-package-hazard-example).
For tree-shaking tests with various bundlers [see here](https://github.com/connectrpc/examples-es/tree/tstamm/dual-package-hazard-example/dual-package-tree-shaking).


timostamm:
@smaye81, can you take a look? From the code size benchmark in connect-es, tree-shaking in esbuild works, but it would be good to have confidence that that's also true for other bundlers like webpack and parcel.

timostamm:
Thanks @smaye81 and @paul-sachs. 

I've moved the tree-shaking tests out to a separate project [here](https://github.com/connectrpc/examples-es/tree/tstamm/dual-package-hazard-example/dual-package-tree-shaking), and added bundlers to the dual package hazard example [here](https://github.com/connectrpc/examples-es/tree/tstamm/dual-package-hazard-example/dual-package-hazard-example). 

The results with the fix from https://github.com/connectrpc/connect-es/pull/842 are:
- no dual package hazard with any consumer (no `FAIL` with `bash test.bash`)
- tree shaking results are exactly the same (no diff after `bash test.bash`)

PR #1068, opened 2023-10-04
Add a cloudflare workers example

srikrsna-buf:
Add a cloudflare workers example.


cloudflare-workers/src/worker-handler.ts
```diff
@@ -0,0 +1,32 @@
+import { createConnectRouter } from '@connectrpc/connect';
+import { universalServerRequestFromFetch, universalServerResponseToFetch } from '@connectrpc/connect/protocol';
+import type { ConnectRouter, ContextValues } from '@connectrpc/connect';
+import type { UniversalHandler } from '@connectrpc/connect/protocol';
+
+/**
+ * Creates new woker handler for the given Connect API routes.
```
timostamm, line 7:
```suggestion
 * Creates new worker handler for the given Connect API routes.
```

cloudflare-workers/src/worker-handler.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+export function createWorkerHandler<Env>(
+	routes: (router: ConnectRouter) => void,
+	options?: {
+		contextValues?: (req: Request, env: Env, ctx: ExecutionContext) => ContextValues;
+		notFound?: (req: Request, env: Env, ctx: ExecutionContext) => Response;
+	}
+) {
```
timostamm, line 15: In other server plugins, we accept a single options object with `routes` that extends the `ConnectRouterOptions`. For example, [express](https://github.com/connectrpc/connect-es/blob/ffd41b00cb8d68d0bbbd08b7faaffc489501b1da/packages/connect-express/src/express-connect-middleware.ts#L31).

I think it makes sense to follow the pattern here.

Putting Connect routes into a separate file makes sense as our general recommendation, but it can also be really convenient to define them in place, and that works out better with `routes` being a property:

```ts
export default createWorkerHandler<Env>({
  contextValues: (_, env) => createContextValues().set(kStore, env.STORE),
  routes(router) {
    router.service(...);
  }
});
```

cloudflare-workers/src/index.ts
```diff
@@ -0,0 +1,36 @@
...
+	//
+	// Example binding to a Queue. Learn more at https://developers.cloudflare.com/queues/javascript-apis/
+	// MY_QUEUE: Queue;
+}
+
+export default {
+	fetch: createWorkerHandler<Env>(routes, {
+		contextValues: (_, env) => createContextValues().set(kStore, env.STORE),
+	}),
+};
```
timostamm, line 36: Should we export an object with a fetch property?

```ts
export default createWorkerHandler<Env>(routes, {
  contextValues: (_, env) => createContextValues().set(kStore, env.STORE),
});
```

It's still possible to get just the `fetch` property if you need it, but it seems like this would be a bit more convenient for the typical use case.

cloudflare-workers/package.json
```diff
@@ -0,0 +1,29 @@
...
+	"version": "0.0.0",
+	"private": true,
+	"type": "module",
+	"scripts": {
+		"deploy": "wrangler deploy",
+		"start": "wrangler dev",
+		"generate": "buf generate ./proto",
+		"build": "wrangler deploy --dry-run --outdir=dist",
+		"test": "node --loader ts-node/esm --no-warnings --test src/url-shortener.test.ts",
+		"ci": "npm run generate && npm run build && npm run test"
```
timostamm, line 12: You'll need to add this project to the [workflow here](https://github.com/connectrpc/examples-es/blob/d3669ab7528439e34cb07c6535a58e18fab338b4/.github/workflows/ci.yaml#L20) so that it runs in CI.

timostamm (commented):
Great practical example 👍 

IMO it's fine to merge this as is and update once we have a server plugin package for cloudflare. I think it's also fine to add it to the top-level README.md, so we don't forget 🙂 

Left just a couple of comments regarding the use-facing API.

timostamm (approved):
LGTM, but let's add this to the top-level readme too.

PR #1075, opened 2023-10-06
Add vercel edge function example

srikrsna-buf:
Add vercel edge function example.

This also adds client usage example, Vercel is [claiming streaming](https://vercel.com/docs/functions/streaming) support. 

Using the fetch client from `connect`, the client for server steaming throws an error at the very end of the steam with: `Invalid state: ReadableStream is locked`.  Which should be fixed in https://github.com/connectrpc/connect-es/pull/876

`connect-web` transport doesn't work on actual because of an unsupported fetch feature. 

`vercel dev` doesn't work because of an issue in `@vercel/node`: https://github.com/vercel/vercel/pull/10701


vercel/edge-functions/api/[service]/[method].ts
```diff
@@ -0,0 +1,58 @@
...
+          };
+        }
+      },
+    });
+  },
+});
+
+export default function POST(req: Request, ctx: RequestContext) {
+  return handler(req, ctx);
+}
```
Comment (timostamm) [inline vercel/edge-functions/api/[service]/[method].ts:58]: Shouldn't we handle Connect GET as well here?

vercel/edge-functions/buf.gen.yaml
```diff
@@ -0,0 +1,9 @@
+# This template is to generate connect-web artifacts for the proto definitions.
```
timostamm, line 1:
```suggestion
# This template is to generate connect-es artifacts for the proto definitions.
```

vercel/edge-functions/lib/functions.ts
```diff
@@ -0,0 +1,95 @@
...
+   */
+  notFound?: (req: Request, ctx: RequestContext) => Promise<Response>;
+  /**
+   * Optional prefix path to strip from the request path before routing.
+   */
+  stripPrefixPath?: string;
+}
+
+/**
+ * Creates new worker handler for the given Connect API routes.
```
timostamm, line 51: Is this a "worker handler"?

vercel/edge-functions/api/[service]/[method].ts
```diff
@@ -0,0 +1,58 @@
...
+import { createContextValues, createPromiseClient } from "@connectrpc/connect";
+import { ElizaService } from "../../lib/gen/connectrpc/eliza/v1/eliza_connect";
+import { createTransport } from "@connectrpc/connect/protocol-connect";
+
+export const config = {
+  runtime: "edge",
+};
+
+const handler = createEdgeFunctionHandler({
+  stripPrefixPath: "/api",
```
Comment (timostamm) [inline vercel/edge-functions/api/[service]/[method].ts:17]: With the Next.js Pages Router, we have to use the "/api" prefix for API routes.

If I remember correctly, this requirement does not exist with the new App Router, and I assume that would be true for a plain edge function as well. 

This might very well be out of scope of this PR, but it's important we know exactly what the constraints about routing are, and that we communicate them clearly.

vercel/edge-functions/api/[service]/[method].ts
```diff
@@ -0,0 +1,58 @@
...
+  contextValues(req) {
+    return createContextValues().set(kGeo, geolocation(req));
+  },
+  routes({ service }) {
+    const eliza = createPromiseClient(
+      ElizaService,
+      createTransport({
```
Comment (timostamm) [inline vercel/edge-functions/api/[service]/[method].ts:24]: This is the transport from `@connectrpc/connect/protocol-connect`. 

The description says "`connect-web` transport is working". So it's possible to use either of them here? If so, can you add a comment?

vercel/edge-functions/lib/functions.ts
```diff
@@ -0,0 +1,95 @@
...
+    };
+    const uRes = await handler(uReq);
+    return universalServerResponseToFetch(uRes);
+  };
+}
+
+/**
+ * Creates a new transport options with the default values.
+ */
+export const defaultTransportOptions = {
```
timostamm, line 86: Yep, that works 😀 

To improve the user experience, we can define a `createConnectTransport` function here that does the necessary plumbing. This might be the best way forward for a future `@connectrpc/connect-vercel` package to make it clear that Connect and gRPC-web are supported, but gRPC is not. 

Lots of open questions. Perhaps we can simply use the `@connectrpc/connect-web` transport in this example for now to a void opening up questions we don't have answers for yet?

vercel/edge-functions/api/[service]/[method].ts
Comment (srikrsna-buf) [reply vercel/edge-functions/api/[service]/[method].ts:24]: Yes, I'll add a comment, although I am not sure if it will work for client streaming which I haven't tried.

srikrsna-buf:
Closing this as stale, will revist after v2

PR #1076, opened 2023-10-07
Add custom client signature example

srikrsna-buf:
The promise client/callback clients are excellent sources for someone looking for guidance on creating custom client signatures. But they can be a bit overwhelming considering the transport interaction and the accompanying test suite. 

This example demonstrates how to change the signature by reusing the promise client.


custom-client/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "buf-custom-client",
+  "version": "0.0.0",
+  "description": "Demonstrates creating custom clients from service definition",
+  "type": "module",
+  "scripts": {
+    "generate": "buf generate buf.build/connectrpc/eliza",
+    "build": "tsc --noEmit",
+    "test": "node --loader ts-node/esm --no-warnings --test src/strict-client.test.ts",
```
timostamm, line 9: Just FYI this also works with `tsx`, which is more lightweight with less dependencies and faster with esbuild. Just have to be aware that `tsx` and `esbuild` does not type-check.

timostamm (approved):

PR #1105, opened 2023-10-30
Update deps

smaye81:
This updates the deps for @bufbuild and @connectrpc to the latest. It also updates the setup-node Action to use the latest LTS.


plain/package.json
```diff
@@ -9,19 +9,19 @@
     "build": "esbuild src/index.ts --bundle --outfile=dist/out.js",
     "typecheck": "tsc --noEmit",
     "generate": "buf generate buf.build/connectrpc/eliza",
-    "test": "NODE_OPTIONS=--experimental-vm-modules ts-node --esm node_modules/.bin/jasmine --config=spec/support/jasmine.json",
+    "test": "node --experimental-loader ts-node/esm node_modules/.bin/jasmine --config=spec/support/jasmine.json",
```
smaye81, line 12: There seems to be a bug with using `ts-node` and its esm loader with Node v20. I was able to confirm this locally. What seems to happen is that an `ERR_UNKNOWN_FILE_EXTENSION` is thrown by `ts-node` and then this causes Jasmine to attempt to fallback to loading the module via `require`, which then ultimately fails with:

```bash
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/sayers/Projects/examples-es/plain/spec/index.spec.ts
require() of ES modules is not supported.
require() of /Users/sayers/Projects/examples-es/plain/spec/index.spec.ts from /Users/sayers/Projects/examples-es/plain/node_modules/jasmine/lib/loader.js is an ES module file as it is a .ts file whose nearest parent package.json contains "type": "module" which defines all .ts files in that package scope as ES modules.
Instead change the requiring code to use import(), or remove "type": "module" from /Users/sayers/Projects/examples-es/plain/package.json.

    at createErrRequireEsm (/Users/sayers/Projects/examples-es/plain/node_modules/ts-node/dist-raw/node-internal-errors.js:46:15)
    at assertScriptCanLoadAsCJSImpl (/Users/sayers/Projects/examples-es/plain/node_modules/ts-node/dist-raw/node-internal-modules-cjs-loader.js:584:11)
    at Object.require.extensions.<computed> [as .ts] (/Users/sayers/Projects/examples-es/plain/node_modules/ts-node/src/index.ts:1610:5)
    at Module.load (node:internal/modules/cjs/loader:1091:32)
    at Function.Module._load (node:internal/modules/cjs/loader:938:12)
    at Module.require (node:internal/modules/cjs/loader:1115:19)
    at require (node:internal/modules/helpers:130:18)
    at Loader.requireShim [as require_] (/Users/sayers/Projects/examples-es/plain/node_modules/jasmine/lib/loader.js:68:10)
    at /Users/sayers/Projects/examples-es/plain/node_modules/jasmine/lib/loader.js:32:27
    at async Jasmine._loadFiles (/Users/sayers/Projects/examples-es/plain/node_modules/jasmine/lib/jasmine.js:142:7) {
  code: 'ERR_REQUIRE_ESM'
}
```

The relevant Jasmine code is [here](https://github.com/jasmine/jasmine-npm/blob/83e59be1d1102e814cc46159e30053077d3648ef/lib/loader.js#L21). The file extension error is thrown by ts-node, which is the bug. This then forces Jasmine to fail. 

I did some searching on ts-node's issues and it seems there is one related to this: https://github.com/TypeStrong/ts-node/issues/1997

So this uses the suggested workaround, which seems to be to use node and instead pass `ts-node/esm` to node as a loader.

timostamm: We're already using `esbuild` - why don't we use `@esbuild-kit/esm-loader` as a loader instead?

Note that it won't check typings, only syntactical errors, same as `esbuild`. Since we have the `typecheck` script, this is just fine.

smaye81: We can, sure. What does this really buy us though? We'd be able to remove `ts-node` but at the expense of adding `@esbuild-kit` as a dep. Plus, it seems like we'd still have to pass it to Node as a loader via the workaround syntax. 

I guess we'd have the benefit of sticking with just esbuild for building and loading and removing a buggy dep?

timostamm: > sticking with just esbuild for building

Yes. Both ts-node and esbuild have limitations. By using both, you get the sum of the limitations.

timostamm: It's `node --loader tsx/esm` now, lo longer `@esbuild-kit/esm-loader`.

smaye81: Looks like it actually changed further than that -- it's `node --import tsx/esm` in 20.6 and beyond. I added an `.nvmrc` to the `plain/` project to enforce the latest v20.

timostamm: Hold on. 

1. `.nvmrc` files don't enforce anything unless the user has a tool installed with shell hooks. 
2. I don't think that `20` will actually enforce the latest version even with a tool installed. I'm quite sure it will accept `v20.0.0` as a match.
3. v20.6.0 was released less than two months ago.

You can't use `node --import tsx/esm` here.

smaye81:
It does enforce it, at least with nvm:

```bash
$ nvm use 20.0.0                                                                                                                                                                                                                                                                                                                                                                  
Now using node v20.0.0 (npm v9.6.4)

$ nvm list                                                                                                                                                                                                                                                                                                                                                                         
       v14.21.3
       v16.15.0
       v16.15.1
       v16.20.0
        v17.9.0
        v18.8.0
       v18.15.0
       v18.16.0
       v18.16.1
       v18.17.0
       v18.17.1
       v18.18.0
        v19.2.0
        v19.3.0
        v19.6.0
->      v20.0.0
        v20.8.0
        v20.9.0

$ cd plain                                                                                                                                                                                                                                                                                                                                                                         Found '/Users/sayers/Projects/examples-es/plain/.nvmrc' with version <20>
Now using node v20.9.0 (npm v10.1.0)

$ cd ..                                                                                                                                                                                                                                                                                                                                                                            

$ nvm use 18                                                                                                                                                                                                                                                                                                                                                                       
Now using node v18.18.0 (npm v9.8.1)

$ cd plain                                                                                                                                                                                                                                                                                                                                                                         Found '/Users/sayers/Projects/examples-es/plain/.nvmrc' with version <20>
Now using node v20.9.0 (npm v10.1.0)
```

Re: using `--import`, we can't use `--loader` either because that doesn't work with 20.6.0 and above. I reverted it back to using `ts-node`. It doesn't seem to make sense to switch it if we have to install `tsx` and using `ts-node` works in all versions.

timostamm:
> It does enforce it, at least with nvm:

You already have v20.9.0 installed, so it's going to pick it for `20`. But if you only have v20.0.0 installed, I expect it to pick this version instead, and not install or ask for the current LTS v20.9.0.

> Re: using --import, we can't use --loader either because that doesn't work with tsx on 20.6.0 and above.

Oh boy :face_with_head_bandage:

timostamm (approved):

PR #1295, opened 2024-01-30
Add RSC example

paul-sachs:
Adding an example of how to work with connect using nextjs RSC and server actions.

This requires that the server actions always communicate with `PlainMessage` types instead of their class based variants. This could be generated as well with a custom plugin if it turns out to be very repetitive but given you often need to validate form data anyways, this may require additional layers to really be useful.


nextjs/app/react-server-actions/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const getMessages = unstable_cache(getMessagesFromCookie, ["my-messages"]);
+
+export default async function Page() {
+  async function submitForm(formData: FormData) {
+    "use server";
+    const sentence = formData.get("chat-message")?.toString() ?? "";
+    addMessageToCookie({ text: sentence, sender: "user" });
+    const response = await say({ sentence });
+    addMessageToCookie({ text: response.sentence, sender: "eliza" });
```
timostamm, line 46: As mentioned above, I believe this is more complex than necessary. We can simply make a regular client call here.

Using cookies is a nice idea, but I think it is a bit distracting from what we want to show. Can we move the get and add functions to a separate file fake-db.tsx? With actions.tsx removed, we'll have page.tsx as the demo, and fake-db.tsx is a less important implementation detail.

nextjs/app/react-server-actions/actions.tsx
```diff
@@ -0,0 +1,26 @@
...
+  })
+);
+
+export type SayResponse = PlainMessage<OrgSayResponse>;
+export type SayRequest = PlainMessage<OrigSayRequest>;
+
+export async function say(content: SayRequest) {
+  const results = await elizaClient.say(content);
+  return toPlainMessage(results);
+}
```
timostamm, line 26: Oof, my comment got lost. Here's the gist:

When the user clicks the button, `FormData` is sent to the server. Since `submitForm()` runs on the server, we don't need `say`, `"use server"`, and plain messages here. We can simply use a regular client in page.tsx instead.

paul-sachs: Unfortunately the `"use server"` is still required. If I don't do that, the compiler complains:

```
 ⨯ Error: Functions cannot be passed directly to Client Components unless you explicitly expose it by marking it with "use server".
  <form action={function} children=...>
               ^^^^^^^^^^
```

I originally split out into the actions file to encapsulate the `toPlainMessage()` wrapping into it's own set of client functions but I can put that into the submitForm function instead.

timostamm: For `submitForm()`, yes. I was referring to `"use server"` in actions.tsx.

paul-sachs: Ah got it.

nextjs/app/react-server-actions/page.tsx
```diff
@@ -0,0 +1,61 @@
...
+        baseUrl: `http://${headers().get("host")}/api`,
+      })
+    );
+
+    const sentence = formData.get("chat-message")?.toString() ?? "";
+    addMessage({ message: { sentence }, sender: "user" });
+
+    const response = await elizaClient.say({ sentence });
+    // toPlainMessage strips the class information from the response, making it safe to pass through the SSR boundary.
+    addMessage({ message: toPlainMessage(response), sender: "eliza" });
```
timostamm, line 29: I don't think you need `toPlainMessage()` here. Everything here runs on the server and does not cross the serialization boundary.

nextjs/app/react-server-actions/fake-db.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+  return [];
+}
+
+export async function addMessage(message: ChatMessage) {
+  const cookie = cookies().get("messages");
+  if (cookie) {
+    const messages = JSON.parse(cookie.value) as ChatMessage[];
+    messages.push(message);
+    cookies().set("messages", JSON.stringify(messages), {
```
timostamm, line 26: JSON.stringify on a PlainMessage is potentially unsafe. If there was a 64-bit integer field, it would crash. I don't think we should use this in an example.

I think we should either store the messages as JSON - but then use the protobuf JSON methods - or store just a hand-written JSON object with the text.

paul-sachs: Oh yeah good call, originally i wasn't storing the full message, and just using a string as text. So yeah changing this makes sense.

paul-sachs: I changed it to store a full message to be a little more realistic of an example.

timostamm: Need to push it up? 🙂

paul-sachs: Ah yes. Not sure i'm super happy with what i've got. But it is just for persistence so it's probably fine.

nextjs/app/react-server-actions/fake-db.tsx
```diff
@@ -0,0 +1,57 @@
...
+  sender: ChatMessage["sender"];
+}
+
+export async function getMessages(): Promise<ChatMessage[]> {
+  const cookie = cookies().get("messages");
+  if (cookie) {
+    const storedMessages = JSON.parse(cookie.value) as StoredChatMessage[];
+    return storedMessages.map((message) => ({
+      sender: message.sender,
+      message: toPlainMessage(
```
timostamm, line 26: toPlainMessage is not necessary here, as explained in https://github.com/connectrpc/examples-es/pull/1295#discussion_r1476488587. 

Let's use either the full classes here, or just a string.

Here is my suggestion for using just a string: https://github.com/connectrpc/examples-es/pull/1364

paul-sachs: Ah i see, yeah that makes sense

timostamm (approved):

PR #1365, opened 2024-02-13
Add bundle-size testing directory

smaye81:
This adds a `bundle-size` directory which assists in testing Connect and Protobuf-ES with various bundlers. It will enable us to test how changes to those libraries affect bundle size and tree-shaking across an array of popular bundlers in the ecosystem.

Note that this directory is not meant to show an example of any sort as the others do. This is just for our own testing purposes.


timostamm:
> @timostamm Let me know what you think about this first attempt at a `bundle-size` directory. It is not yet currently wired up to CI because I wanted to get your thoughts first.
> 
> Also, it is not yet hooked into `manage.mjs`. I think it should at least be a part of the `forceupdate` command so that we can easily bump up the dependencies, but wasn't sure if you thought it should be a part of the test run of that script also.

manage.mjs expects the scripts "generate", "build", "test", and "ci". https://github.com/connectrpc/examples-es/pull/1366 adds them. This means "bundle-size" can now use mange.mjs's commands, and should also plug nicely into ci (you need to update the workflow).

bundle-size/consumer-parcel/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "@bufbuild/buf": "^1.29.0",
+    "@bufbuild/protoc-gen-es": "^1.7.2",
+    "@connectrpc/protoc-gen-connect-es": "^1.3.0"
+  },
+  "//": "otherwise, parcel shakes everything away",
+  "@parcel/resolver-default": {
+    "packageExports": true
+  },
+  "source": "src/index.html",
+  "sideEffects": true
```
timostamm, line 25: Please take a look why we set this property, also elsewhere. https://github.com/connectrpc/examples-es/pull/1366#discussion_r1488146329

bundle-size/consumer-esbuild/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "consumer-esbuild",
```
timostamm, line 2: consumer-* doesn't roll off the tongue. But we should avoid using the same package name as the upstream npm package. WDYT about renaming just the directories, removing the "consumer-" prefix, and keeping the package names as is?

bundle-size/README.md
```diff
@@ -0,0 +1,70 @@
...
+The results are:
+
+<!--- RESULTS-START -->
+| Name | Size | Artifact |
+|------|-----:|----------|
+| esbuild | 82,796 bytes | dist/index.js |
+| webpack | 33,817 bytes | dist/main.js |
+| webpack-ts | 33,817 bytes | dist/bundle.js |
+| vite | 54,732 bytes | dist/consumer-vite.js |
+| rollup | 145,459 bytes | dist/index.js |
```
timostamm, line 24: This does look like it could be https://github.com/bufbuild/protobuf-es/pull/470 🤔

timostamm (commented):
Looks good to me, just two nits:

bundle-size/consumer-esbuild/package.json
smaye81: Yep, I didn't like the `consumer-` prefix either.

bundle-size/consumer-parcel/package.json
smaye81: OK, I did some research and testing:

* Parcel - supports `sideEffects` and defaults to `true` if absent. Explicitly setting it to false makes about a 60B difference in our bundle size test for Parcel.
* esbuild - not supported / has no effect
* Rollup - not supported / has no effect. Rollup has its [own config](https://rollupjs.org/configuration-options/#treeshake-modulesideeffects) for side effects, but worth noting it has no effect on our bundle size tests whether it's set or not.
* Vite - not supported / has no effect
* Webpack - supports `sideEffects` but has no effect on our bundle size tests whether it's set or not.

Based on the above, I removed `sideEffects` from all package.json files, except for Parcel where I left it in and set it to false.

timostamm: Parcel removes the console.log statements from index.js with sideEffects: false. Seems a bit drastic to me 🤷

Let's just remove the sideEffects property for Parcel as well.

smaye81: Agreed. Removed.

timostamm (approved):
LGTM!

PR #1366, opened 2024-02-13
Switch bundle-size to use npm scripts

timostamm:


bundle-size/consumer-parcel/package.json
```diff
@@ -16,9 +22,5 @@
     "packageExports": true
   },
   "source": "src/index.html",
-  "sideEffects": true,
-  "scripts": {
-    "start": "rm -rf dist && npx buf generate buf.build/connectrpc/eliza && parcel build --log-level error --no-cache --no-content-hash"
-  },
-  "testOutputPath": "dist/index.bfe2d88b.js"
+  "sideEffects": true
```
smaye81, line 25: Just noticed this is the only one that's `true`. Should this also be `false`?

timostamm: I only updated the scripts, and didn't look at anything else. It does look wrong. We should try to remove the property (should be equivalent to sideEffects:true).

PR #1444, opened 2024-03-12
Add additional testing with bundle sizes for TypeScript

smaye81:
This expands the `bundle-size` directory to include testing and size calculations for both JavaScript and TypeScript. Currently, the code only calculates bundle size for JavaScript and generated JS files from Protobuf-ES and it does not verify that the code generated by the bundlers actually runs.

Now, each subdirectory for the bundlers has been expanded to generate JS files as well as TS+DTS files. The bundlers bundle each as separate entry points and then calculate the bundle size for TypeScript as well as JavaScript. In addition, it runs the code generated by the bundlers for both TS and JS and verifies that imports are correct and the code runs.

This is helpful to test code that returns type information at runtime. For example, it allows us to now verify that adding `PURE` annotations to generated TypeScript for enums will throw errors at runtime.


smaye81:
@timostamm Verified that the `PURE` annotations work correctly in JS-generated code. In addition, as mentioned in the description, I verified that adding `PURE` to `setEnumType` in the TypeScript-generated code failed at runtime when trying to call `getEnumType`.

timostamm:
> Verified that the `PURE` annotations work correctly in JS-generated code. In addition, as mentioned in the description, I verified that adding `PURE` to `setEnumType` in the TypeScript-generated code failed at runtime when trying to call `getEnumType`.

Thanks! Very useful to have this confirmed. Looks like we're now good to merge https://github.com/bufbuild/protobuf-es/pull/470.

timostamm (approved):

PR #1488, opened 2024-04-03
Update dependabot config with new directories

smaye81:
This updates the dependabot.yml config to add some directories that were missing / incorrect.

Note:
- Since we have the `manage.mjs` script, it is not totally necessary to have every directory in dependabot.yml, but keeping it in sync has two benefits: 1. it cross-checks that our `manage.mjs` script is accurately updating all dependencies in the repo (otherwise those dependabot PRs will stay open). 2. For completeness sake since it feels odd to have only some of our directories in that config.
- The `bundle-size` has also been added and each bundler is its own group. Currently, the `manage.mjs` script does not handle workspaces, which is how the `bundle-size` directory is structured, so we actually _need_ that directory to be managed by dependabot. We could update `manage.mjs` to handle the workspace setup in the future if need be.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,6 +18,102 @@ updates:
       timezone: UTC
       time: "07:00"
   - package-ecosystem: "npm"
+    directory: "astro"
+    schedule:
+      interval: "monthly"
+      timezone: UTC
+      time: "07:00"
+  - package-ecosystem: "npm"
+    directory: "bundle-size"
+    schedule:
+      interval: "monthly"
+      timezone: UTC
+      time: "07:00"
+  - package-ecosystem: "npm"
+    directory: "bundle-size/esbuild"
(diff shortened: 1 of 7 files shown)


timostamm (commented):
I expect manage.mjs's `update` to run `npm update`, which updates within constraints. 
`list`, `test`, and `ci` should work as well, so that CI is covered.

I expect `forceupdateknown` and `forceupdateall` not to work for the workspace dependencies. So if one of esbuild, parcel, etc. has a major new version (or a minor for esbuild, since it's pre-release), they won't be updated by the script.

Is this what you saw?

timostamm:
It seems worth trying to update the script. If it turns out to be a simple change, it'll definitely help with maintenance. 

LGTM though, it makes sense to update the missing dependabot directories regardless of the script.

PR #1490, opened 2024-04-04
Enhance manage script for npm workspaces

smaye81:
This adds some additional functionality to the `manage.mjs` script to handle npm workspaces. Currently, the `bundle-size` directory is somewhat of an anomaly in the repo. It's main purpose is testing and is structured using npm workspaces. This meant that the previous incarnation of the manage script would not perform the necessary package.json updates when `forceupdateknown` and `forceupdateall` were called. 

This modifies the script so that any package that contains workspaces maintains the workspaces as part of the `PackageEnt` class and knows how to recursively call the workspace update functions.


bundle-size/README.md
```diff
@@ -17,16 +17,16 @@ The results are:
...
-| parcel (ts) | 76,825 bytes | dist/ts/index.dc635180.js |
-| rollup (js) | 145,627 bytes | dist/js/index.js |
-| rollup (ts) | 155,601 bytes | dist/ts/index.js |
-| vite (js) | 55,512 bytes | dist/js/consumer-vite.js |
-| vite (ts) | 58,166 bytes | dist/ts/consumer-vite.js |
-| webpack (js) | 34,474 bytes | dist/js/main.js |
-| webpack (ts) | 37,435 bytes | dist/ts/main.js |
+| esbuild (js) | 83,451 bytes | dist/js/index.js |
+| esbuild (ts) | 89,967 bytes | dist/ts/index.js |
+| parcel (js) | 37,145 bytes | dist/js/index.cf4251ac.js |
```
smaye81, line 22: [Parcel's v2.10](https://github.com/parcel-bundler/parcel/releases/tag/v2.10.0) claims to include reduced runtime bundle sizes, which I'm guessing is why the drastic reduction in size here.

manage.mjs
```diff
@@ -152,6 +147,15 @@ class PackageEnt {
         this.name = pkgJson.name;
         this.packageJson = pkgJson;
         this.packageManager = packageManager;
+        this.workspaces = [];
+        // If this is a package with workspaces, create a PackageEnt for each
+        // and store it as part of the main package
+        if (pkgJson.workspaces) {
+            this.workspaces = pkgJson.workspaces.map((ws) => {
+                const pkgPath = path.join(dir, ws, "package.json");
```
smaye81, line 155: Note I'm not checking whether `pkgPath` exists as `tryGetPackage` does. Assuming here that since it's part of a workspace, the package.json should exist.

manage.mjs
```diff
@@ -191,7 +195,16 @@ class PackageEnt {
...
     /**
+     * @param {number} indent - Amount to indent the output. Used for workspaces.
      */
+    print(indent = 0) {
+        const spacing = ' '.repeat(indent);
+        console.log(`${spacing}${this.name} (${this.packageManager}) at ${this.path}`);
+        for (const ws of this.workspaces) {
+            ws.print(2);
+        }
+    }
```
timostamm, line 206:
```suggestion
    toString() {
        const ws = this.workspaces.length > 0 ? `, ${this.workspaces.length} workspaces` : "";
        return `${this.name} (${this.packageManager}${ws}) at ${this.path}`;
    }
```

Will print:
```
bundle-size (npm, 5 workspaces) at /.../examples-es/bundle-size
```

I think this is a better fit than listing the individual workspaces, because `manage test` and other commands don't let you specify a workspace package, only the root package.

(Needs `console.log(pkg.toString())` in `main()`)

bundle-size/README.md
timostamm: Isn't that rather because protobuf-es v1.8.0 adds PURE annotations? 

Then again, `parcel (ts)` also shrinks 🤔

Let's update dependencies separately.

I noticed that we generate .ts and .d.ts files (which doesn't make much sense, it should only be .ts), and standalone .js files (they should have the .d.ts file next to them). We need to fix this. Probably also best separately.

timostamm (approved):

PR #1499, opened 2024-04-10
Treeshaking

smaye81:


smaye81:
@timostamm here's an example of trying the `PURE` annotation with `newFieldList`. It seems to make a difference for esbuild, but no other bundlers. I dug into some of the numbers for the other bundlers and  I _think_ this is what's happening:

### Parcel
Parcel doesn't seem to be tree-shaking _any_ code from `node_modules`. If you search for an unused message like `SayResponse` in the output bundle, you'll see it's included there.

Also, Parcel's numbers aren't affected at all because the output is minified by default and the comments are stripped away. You can turn it off with `--no-optimize`, but that also turns off tree-shaking 😞 

### Rollup
The `PURE` annotation for `newFieldList` doesn't have any effect. I tried a few different configurations in rollup.config to see if it would help, but to no avail. Rollup's numbers aren't affected at all because the `PURE` annotation is being stripped in the output bundle. 

Rollup [claims](https://rollupjs.org/configuration-options/#pure) that it removes invalid annotations and emits a warning, but I don't see any warning in the build (with `silent` off). The best I can guess is that the TypeScript plugin for Rollup is causing problems and might be stripping this annotation?

### Vite
Like Rollup, the `PURE` annotation for `newFieldList` doesn't have any effect. Vite's numbers are actually inflated because the `PURE` annotation is appearing in the output bundle. I also looked into why Vite's output is so much smaller than Rollup's and I believe it's because Vite is doing additional mangling to reduce the size whereas Rollup is just straight bundling. 

### Webpack
Like Rollup and Vite, the `PURE` annotation for `newFieldList` doesn't have any affect. Webpack's numbers aren't affected either because Webpack is minifying the output and stripping the annotation.

liril-net:
Hi @smaye81 , i think it not works because now the `fields` call `proto3.util.newFieldList` to init its value, this makes it always `run` when the file is imported, for my personal fork, i make it as  `fields: () => proto3.util.newFieldList()` to ensure it as a lazy init and it will works for `tree-shaking`.

smaye81:
Ah right. Thanks for the clarification. I don't think we will be making that lazy-initialization refactor since that would be breaking change to generated code. We are actually working on a v2 of Protobuf-ES and plan to release an alpha version shortly, so bundle sizes, etc. will be a bit in flux. We can revisit once we get a better benchmark with the v2 release.

In the meantime, let's keep https://github.com/bufbuild/protobuf-es/issues/769 open just for tracking so that we can verify when the release is ready.

liril-net:
cool, where can i see the v2 working progress?

smaye81:
Work is being done on the `v2` branch in the https://github.com/bufbuild/protobuf-es repo.

timostamm:
The `@__PURE__` annotation should inform Rollup that the call is free of side-effects, and that it's safe to shake other parts of this module. This might be a shortcoming in Rollup's tree-shaking.

PR #1566, opened 2024-05-01
Allow skipping packages during processes

smaye81:
There are occasions when updating dependencies for a certain project fail, which in turn, fails the entire process. For example, today (5/1), updating the `react-native` project failed as a result of peer dependencies. To get the rest of the projects to update, `react-native` had to be manually skipped.

This allows for skipping packages when running a process by prefixing it with `!`. For example:

`./manage.mjs list '!buf-react-native'`

### DIFF
--- a/manage.mjs
+++ b/manage.mjs
@@ -38,10 +38,13 @@ function main() {
         onlyPackages.push(pkg);
       }
     });
+    if (onlyPackages.length > 0 && notPackages.length > 0) {
+      process.stderr.write("either a list of packages or a list of skipped packages must be provided, but not both");
+      process.exit(1);
+    }
     if (onlyPackages.length > 0) {
         packages = packages.filter(p => onlyPackages.includes(p.name));
-    }
-    if (notPackages.length > 0) {
+    } else if (notPackages.length > 0) {
         packages = packages.filter(p => !notPackages.includes(p.name));
     }
     switch (command) {
--- a/manage.mjs
+++ b/manage.mjs
@@ -28,10 +28,21 @@ const knownDependencies = [
 
 function main() {
     const command = process.argv[2];
-    const onlyPackages = process.argv.splice(3);
+    const packageList = process.argv.splice(3);
     let packages = listPackages(process.cwd());
+    let onlyPackages = [], notPackages = [];
+    packageList.forEach((pkg) => {
+      if (pkg.startsWith("!")) {
+        notPackages.push(pkg.substring(1));
+      } else {
+        onlyPackages.push(pkg);
+      }
+    });
     if (onlyPackages.length > 0) {
         packages = packages.filter(p => onlyPackages.includes(p.name));
+    }
+    if (notPackages.length > 0) {
+        packages = packages.filter(p => !notPackages.includes(p.name));
     }
     switch (command) {
         case "list":
--- a/manage.mjs
+++ b/manage.mjs
@@ -32,17 +32,17 @@ function main() {
     let packages = listPackages(process.cwd());
     let onlyPackages = [], notPackages = [];
     packageList.forEach((pkg) => {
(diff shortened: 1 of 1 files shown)


timostamm:
Ok, so `npm update` fails to update react to 18.3.1 because react-native has a peer dependency constraint on react 18.2.0.

The correct solution here is to change the dependency constraints:

```diff
-    "react": "^18.2.0",
+    "react": "18.2.0",
-    "react-dom": "^18.2.0",
+    "react-dom": "18.2.0",
```

Anyhow, when I try to run the project, it fails:

```
$ cd react-native
$ npm ci
$ npm start
npm ERR! Missing script: "buf:generate"
```

After fixing the script, Expo starts up, but I get a whole lot of build errors. react-native is not stable - just running npm updates is not enough, and the example is broken.

It looks like they started using an [upgrade tool](https://react-native-community.github.io/upgrade-helper/?from=0.73.8&to=0.74.0). Running the suggested command:

```
$ npx @rnx-kit/align-deps --requirements react-native@0.74
warn Known bad packages are found in 'package.json':
        metro-react-native-babel-preset@*: This package was renamed to '@react-native/babel-preset' in react-native 0.73. Replace this package when you're on react-native 0.73 or higher.
error package.json: Found 2 violation(s) outside of capabilities.
        In dependencies:
          - react "^18.2.0" should be "18.2.0"
        In devDependencies:
          - jest "^29.7.0" should be "^29.2.1"
error Re-run with '--write' to fix them.
```

I don't think adding a feature to ignore the project is helping us with this. In case of an error like this, it's always worth taking a closer look at what's actually going on.

Going forward, we have to either:
- fix the react-native example for the recent versions
- revert it to a working version and pin it
- delete it from the repository

smaye81:
That's fine, but if we decide on one of the first two options, we still can't run our `manage.mjs` script against the `react-native` directory. Since it removes all deps, it always tries to reinstall latest deps and if there's any issues, it blocks the entire process from completing.

I think once we fix the directory (see https://github.com/connectrpc/examples-es/pull/1567), we also need a way to handle this with the `manage.mjs` script. So we either need to skip the `react-native` directory somehow or maybe if an `npm install` fails for any directory, we revert that directory and move on to the next.

timostamm:
>  it always tries to reinstall latest deps

Only with the `forceupdateall` command, and reaching to that command first thing is not good idea. It's not mentioned in the [guide for updating](https://github.com/connectrpc/examples-es/blob/71f640aae84f8b5b5c9ba8be56c15a49f6f2f30c/.github/CONTRIBUTING.md#updating-dependencies) because it's a footgun.  I would not expect [Angular](https://update.angular.io/?v=16.0-17.0), [React](https://react.dev/blog/2022/03/08/react-18-upgrade-guide), [Svelte](https://svelte.dev/docs/v4-migration-guide), [Remix](https://remix.run/docs/en/main/start/v2), [Astro](https://docs.astro.build/en/guides/upgrade-to/v4/) and other frameworks to be able to upgrade across major versions automatically. Even if builds and tests pass, the app might not work as intended anymore.

I know it's tempting to use `forceupdateall` because it also updates package.json, which makes it possible to see what's been updated. But it doesn't seem right to make the footgun more convenient.

I'd rather we make it more safe. For example:
- skip pinned versions (e.g. `"react": "18.2.0"` in react-native) and alert the user
- alert the user of breaking changes (e.g. going from x@1.2.3 to x@2.0.0, but also going from y@0.1.2 to y@0.2.0)

If we print out the updates with potential breaking changes, it's much easier to take a closer look, because it will only be a handful at most. Then we can make `forceupdateall` the preferred command for updates, because it reduces busywork for maintenance, but still alerts to potential problems that require manual checks. Then it also makes sense to have the capability to skip select projects.

PR #1569, opened 2024-05-03
Add better logic around dependency updates

smaye81:
This adds logic to the `manage.mjs` script which does a few things to mitigate issues of conflicting dependencies:

* It warns and skips installing any pinned dependencies
* It warns if a dependency update could be considered breaking
* It warns if an unrecognized dependency is found

The script prints the warnings as the upgrade is running, but it will also print a report of the above issues once the update is done:

image


smaye81:
@timostamm Tested and it seems to work fine. A few minor things I saw and addressed in https://github.com/connectrpc/examples-es/pull/1569/commits/ed2f9214642fa60292b7e8252890ce35b18f4d9f.

manage.mjs
```diff
@@ -122,6 +89,42 @@ function tryGetPackage(dir) {
...
+    console.log('\x1b[33m%s\x1b[0m', `WARNING: ${text}`);
+}
+
+class UpgradeStats {
+    /**
+     * @param {PackageEnt} pkg
+     * @param {string} dependency
+     */
+    skipPinned(pkg, dependency) {
+      info(`Skipping upgrade of pinned dependency ${dependency} for ${pkg.toString()}.`);
```
timostamm, line 105: Let's print the warnings (in case an install fails, we'll never reach the end of the loop on line 31), but also keep them in memory to print as the last piece of output (the output is very long, and the warnings will disappear within it).

manage.mjs
```diff
@@ -122,6 +89,42 @@ function tryGetPackage(dir) {
     return new PackageEnt(pkgPath);
 }
 
+function info(text) {
+    console.log('\x1b[36m%s\x1b[0m', `INFO: ${text}`);
+}
+function warn(text) {
+    console.log('\x1b[33m%s\x1b[0m', `WARNING: ${text}`);
+}
```
timostamm, line 97: Let's use warnings only, no info. Let's add this as a method on UpgradeStats.

manage.mjs
```diff
@@ -291,15 +297,84 @@ class PackageEnt {
...
+                    }
+                    continue;
+                }
+            }
+            stats.unrecognized(this, key, oldConstraint, newConstraint);
+        }
+
         // Loop through any workspaces and call their forceupdate
-        // Note that forceUpdate removes deps first and then re-installs so packages
+        // Note that upgrade() removes deps first and then re-installs so packages
```
timostamm, line 329: Line 329: /s/forceupdate/upgrade

manage.mjs
```diff
@@ -69,13 +37,12 @@ function main() {
...
-            console.error("'list' lists all packages in the repository.");
-            console.error("'test' installs dependencies, then runs the 'generate', 'build', and 'test' scripts.");
-            console.error("'update' updates all deps to the latest version allowed by the dependency constraints.");
-            console.error("'forceupdateknown' updates all known deps to the latest version, regardless of constraints.");
-            console.error("'forceupdateall' updates all deps to the latest version, regardless of constraints.");
-            console.error("If no packages are given, the command runs for all packages.");
+            console.error("Usage: (list | test | upgrade | update) <projects>...");
+            console.error("'list' lists all projects in the repository.");
+            console.error("'ci' installs dependencies, then runs the script 'ci' for each project.");
+            console.error("'upgrade' updates all dependencies to their latest version, for each project.");
```
timostamm, line 43:
```suggestion
            console.error("'upgrade' updates all dependencies to their latest version, for each project. The command may print warnings at the end, for example when a major upgrade was installed, and you need to look out for breaking changes.");
```

Can you update "Updating dependencies" in contributing.md?

Best to call out that it's usually necessary to follow the framework's upgrade guide for major releases, and link to the examples given in https://github.com/connectrpc/examples-es/pull/1566#issuecomment-2093205231 and to the react native upgrade tool.

smaye81:
@timostamm addressed feedback and added additional logic to print the summary of issues found at the end.

manage.mjs
```diff
@@ -123,6 +91,109 @@ function tryGetPackage(dir) {
...
+
+    statTypes = ["skipped", "breaking", "unrecognized"];
+    summary = {};
+
+    /**
+     * @param {string} statType
+     * @param {string} pkg
+     * @param {string} dependency
+     * @param {string} oldConstraint
+     * @param {string} newConstraint
```
timostamm, line 104:
```suggestion
     * @param {string} [oldConstraint]
     * @param {string} [newConstraint]
```

manage.mjs
```diff
@@ -123,6 +91,109 @@ function tryGetPackage(dir) {
...
+     */
+    unrecognized(pkg, dependency, oldConstraint, newConstraint) {
+        this.cache("unrecognized", pkg, dependency, oldConstraint, newConstraint);
+        this.warn(`Found unrecognized dependency ${dependency} for ${pkg.toString()} while trying to upgrade from v${oldConstraint} to v${newConstraint}.`)
+    }
+
+    /**
+     * @param {string} text
+     */
+    warn(text) {
```
timostamm, line 192:
```suggestion
    #warn(text) {
```

manage.mjs
```diff
@@ -123,6 +91,109 @@ function tryGetPackage(dir) {
 }
 
 
+class UpgradeStats {
+
+    statTypes = ["skipped", "breaking", "unrecognized"];
```
timostamm, line 96:
```suggestion
    #statTypes = ["skipped", "breaking", "unrecognized"];
```

manage.mjs
```diff
@@ -123,6 +91,109 @@ function tryGetPackage(dir) {
...
+    summary = {};
```
timostamm, line 97:
```suggestion
    #summary = {};
```

A nested map holding arrays of objects, that's gutsy without type safety or test coverage.

I think it's better to store in a flat list, and only group when when printing.

manage.mjs
```diff
@@ -123,6 +91,109 @@ function tryGetPackage(dir) {
...
+
+    /**
+     * @param {string} statType
+     * @param {string} pkg
```
timostamm, line 101: Wait, are we passing a `string`, or a `PackageEnt`? I think you're implicitly relying on PackageEnt.toString here, but I'm not sure I understand it?

smaye81: It's a string and in fact, the jsdoc for the other methods is incorrect. They list a `PackageEnt`, but it's a string being passed to all of them. I've updated the doc and the arg name to make it more clear.

manage.mjs
```diff
@@ -123,6 +91,109 @@ function tryGetPackage(dir) {
...
+     */
+    #warn(text) {
+        console.log('\x1b[33m%s\x1b[0m', text);
+    }
+
+    /**
+     * @param {string} pkgName
+     * @param {string} dependency
+     */
+    skipPinned(pkgName, dependency) {
```
timostamm, line 169: `pkgName` should be a string according to the JSDoc, but you're passing in a `PackageEnt`.

```suggestion
    skipPinned(pkgName, dependency) {
      if (typeof pkgName != "string") throw new Error("oops");
```

```bash
$ ./manage.mjs upgrade buf-react-native
Error: oops
```

Best to pass `PackageEnt` around, since it's the more concrete type.

timostamm (approved):
LGTM!

PR #1574, opened 2024-05-15
Update react-native dependencies and layout

### LINKED ISSUE #1573: React Native Example start error.
Hello,

When trying to open the react native example in either web, android, or ios, I get hit with an error.

> ERROR in ./node_modules/expo/build/launch/withDevTools.web.js
> Module build failed (from ./node_modules/babel-loader/lib/index.js):
> SyntaxError: /Users/giovannicoakley/Documents/demo/connectrpc examples-es main react-native/node_modules/expo/build/launch/withDevTools.web.js: Duplicate __self prop found. You are most likely using the deprecated transform-react-jsx-self Babel plugin. Both __source and __self are automatically set when using the automatic runtime. Please remove transform-react-jsx-source and transform-react-jsx-self from your Babel config.
>   4 |     function WithDevTools(props) {
>   5 |         return (<>
>  ->6 |         <AppRootComponent {...props}/>
>     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   7 |         <DevLoadingView />
>   8 |       </>);
>   9 |     }
>     at File.buildError [as buildCodeFrameError] (/Users/giovannicoakley/Documents/demo/connectrpc examples-es main react-native/node_modules/@babel/core/src/transformation/file/file.ts:249:12)
> 

This is after downloading the repo, running npm install and then npm start.

smaye81:
Fixes #1573 

This updates the `react-native` project in the repo to use latest React Native dependencies. To do so, we simply started the project from scratch using latest version of Expo and then rebuilt from the older project.


react-native/app/custom-transport.ts
```diff
@@ -187,12 +187,12 @@ export function createXHRGrpcWebTransport(
...
           }
 
-          return <UnaryResponse<I, O>>{
+          return {
             stream: false,
             header: response.headers,
             message,
             trailer,
-          };
+          } as UnaryResponse<I, O>;
```
timostamm, line 195: You can use `satisfies` instead of `as` here.

timostamm (approved):

PR #1708, opened 2024-07-09
Fix polyfills using expo-dev-client and platform splitting

ryanlntn:
Fixes the RN polyfill updates by:

1. Updating the app to use `expo-dev-client`. This was required as `react-native-fast-encoder` is a native JSI implementation and not provided in expo go.
2. Platform splitting the polyfills. The RN polyfill function is unfortunately a deep import and causes us to run into [errors](https://github.com/expo/fyi/blob/main/fb-batched-bridge-config-web.md) on web unless we completely remove that code from the web bundle.

Demo:
Screenshot 2024-07-09 at 4 16 44 PM

### DIFF
--- a/react-native/.gitignore
+++ b/react-native/.gitignore
@@ -9,6 +9,8 @@ npm-debug.*
 *.mobileprovision
 *.orig.*
 web-build/
+android
+ios
 
 # macOS
 .DS_Store
--- a/react-native/app.json
+++ b/react-native/app.json
@@ -13,21 +13,19 @@
       "backgroundColor": "#ffffff"
     },
     "ios": {
+      "bundleIdentifier": "com.eliza.app",
       "supportsTablet": true
     },
     "android": {
       "adaptiveIcon": {
         "foregroundImage": "./assets/images/adaptive-icon.png",
         "backgroundColor": "#ffffff"
-      }
+      },
+      "package": "com.eliza.app"
     },
     "web": {
-      "bundler": "metro",
-      "output": "static",
-      "favicon": "./assets/images/favicon.png"
+      "bundler": "metro"
     },
-    "plugins": [
-      "expo-router"
-    ]
+    "plugins": ["expo-router"]
   }
 }
--- a/react-native/app/polyfills.native.ts
+++ b/react-native/app/polyfills.native.ts
@@ -1,0 +1,20 @@
+import TextEncoder from "react-native-fast-encoder";
+// @ts-expect-error -- missing type declarations
+import { polyfillGlobal } from "react-native/Libraries/Utilities/PolyfillFunctions";
+// @ts-expect-error -- missing type declarations
+import { fetch, Headers, Request, Response } from "react-native-fetch-api";
+import { ReadableStream } from "web-streams-polyfill";
+
+export function polyfills() {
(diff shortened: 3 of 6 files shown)


smaye81:
Thanks for putting this up! Left a small comment about docs. Two other things I've noticed:

1. Can we change the `start` command to default opening in Expo Go? On main, it currently defaults to that, which allows you to scan the QR code and get up and running on a device. But for some reason, the development build is the default now when running `start`. I'm guessing this changed in a later version? I think just changing the `start` command to be `"start": "npm run generate && expo start --go"` should handle it.

2. Related to running on a device, I still get the same error that I was seeing when trying this the other day. I'm not sure if this is my setup or an actual polyfill bug. Can you test to see if this fails for you also? This actually works on main, so something must've changed. The error I see is:

image

ryanlntn:
@smaye81 

> 1. Can we change the `start` command to default opening in Expo Go? 

We can but you will run into the error from point 2 whenever you attempt to use the app in Expo Go since it doesn't include the react-native-fast-encoder native code and TextEncoder/TextDecoder will throw when you attempt to serialize/deserialize.

> 2. Related to running on a device, I still get the same error that I was seeing when trying this the other day. I'm not sure if this is my setup or an actual polyfill bug. Can you test to see if this fails for you also?

Are you seeing this running Expo Go or the Expo Dev Client? Only the later will include the react-native-fast-encoder native code. When running the Expo Dev Client it works for me.

If we want to be able to continue using Expo Go we could switch back to using [fast-text-encoding](https://github.com/samthor/fast-text-encoding) since that's a non-native implementation (also significantly less performant). Expo also [just added their own implementation of TextDecoder](https://github.com/expo/expo/pull/29620) that would remove the need for react-native-fast-encoder but it hasn't yet landed in a release.

smaye81:
Ah gotcha. That makes sense. Thanks for the explanation. Two questions:

1. Is there any way to get this to work on Expo Go _and_ development builds? Or do we have to pick one vs. the other (and consequently then we'd have to pick between `react-native-fast-encoder` and `fast-text-encoding`)? If it has to be one or the other that's fine, we should just probably document that somewhere in the README.

2. Does scanning the QR code not work at all when using development builds? I get a message on iOS that says 'No usable data found' despite it saying `› Scan the QR code above to open the project in a development build.`

ryanlntn:
> 1. Is there any way to get this to work on Expo Go _and_ development builds? Or do we have to pick one vs. the other (and consequently then we'd have to pick between `react-native-fast-encoder` and `fast-text-encoding`)? If it has to be one or the other that's fine, we should just probably document that somewhere in the README.

@smaye81 We could include both polyfills and apply `fast-text-encoding` if we're in Expo Go or `react-native-fast-encoder` if we're not. I can update to do that if that makes sense.

> 2. Does scanning the QR code not work at all when using development builds? I get a message on iOS that says 'No usable data found' despite it saying `› Scan the QR code above to open the project in a development build.`

The dev build would need to be installed on your device and your device would need to be connected to the same network I believe. The QR code is less useful with dev client builds since you're typically building directly to simulator or device anyway and not just loading a different JS bundle from metro.

smaye81:
Looks great @ryanlntn. Thank you for doing this!

PR #1718, opened 2024-07-29
Upgrade nextjs example to NextJS 14

smaye81:
This updates the example in `nextjs` to use Next.js v14. A few minor changes to accommodate v14:

- Accessing dynamic data sources such as cookies is not permitted inside a cache scope, so the `getMessages` function we use for a fake DB had to be refactored to accept a cookie instead of doing the lookup inside: See https://nextjs.org/docs/app/api-reference/functions/unstable_cache
- `serverActions` are no longer experimental and are the default so that was removed from the nextjs config file.


smaye81:
Once connectrpc/connect-es#1159 is merged and released, we can upgrade these examples to use the new peer deps.

timostamm (commented):
Can you check that we're using Next.js like intended?

IIRC v14 adds a path alias to tsconfig.json. We should use @-imports if that's what Next.js defaults to.

From what I remember, v14 also uses an mjs file for the config, and next-env.d.ts should be git-ignored (we should be using the .gitignore Next.js scaffolds).

smaye81:
Regenerated a Next.js app using v14 and all suggested defaults. Hopefully this should be adhering to v14 patterns.

timostamm:
Looks great! Keeping the scaffolded files up-to-date is super helpful for users setting up a project, and it gives us confidence that we're compatible.

I don't think we should use Tailwind though 😬

I didn't realize that `create-next-app` saved my settings (can delete them with `npx create-next-app@latest --reset-preferences`) and though it wasn't a default choice. 

Tailwind is actually fine, but we shouldn't have Tailwind on board and then use CSS modules instead. I guess we opted for CSS modules because it's much more simple, and we don't care about CSS in our examples.

So we should either remove Tailwind, or port the styles to use Tailwind.

smaye81:
Removed Tailwind (and PostCSS) in favor of the previous CSS modules we've always used. Also updated the README to reflect that we opted out of Tailwind.

timostamm:
Nice, let's release the upstream update and get this merged.

timostamm (approved):

PR #1722, opened 2024-07-30
Fix Vite CJS build warning for Yarn and Vue

### LINKED ISSUE #1721: Avoid Vite's deprecated CJS build of the Node API
I noticed the following warning in react/yarn-pnp:

> The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.

This happens with `yarn run build`.

smaye81:
Fixes #1721 

This fixes the warning about Vite's deprecated CJS build in the `react/yarn-pnp` and `vue` projects by converting them to an ESM module. Previously, they were both CJS, most likely due to a simple oversight.

With Vue, it also involved renaming the Cypress config to be a `.js` file. Vue scaffolds Cypress config files to be a `.ts` file, which in turn does not work with ESM modules. See [this issue](https://github.com/cypress-io/cypress/issues/23552) for more info.

For more info on the Vite CJS message: https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated


timostamm (approved):

timostamm:
Just noticed: The warning also appears for `vue`.

vue/cypress.config.js
```diff
@@ -0,0 +1,42 @@
+import { defineConfig } from "cypress";
+import webpackPreprocessor from "@cypress/webpack-preprocessor";
+import ResolveTypeScriptPlugin from "resolve-typescript-plugin";
+
+// Note that vue scaffolds a cypress config with a .ts extension, but this doesn't
+// work with ESM modules. Renaming this file to have a .js extension fixes the issue.
+// See: https://github.com/cypress-io/cypress/issues/23552
```
timostamm, line 7: `npm create vue@latest` actually does scaffold a .ts file, and it doesn't use the webpack-preprocessor.

I think we might only need the webpack-preprocessor for .js import extensions. We don't add them with v2 anymore, so I think we can simplify this example a lot, and reduce the diff to the default setup. But I don't think it's worth doing that here, in v1.

timostamm (approved):

PR #1787, opened 2024-08-01
Update repo deps for August 2024

smaye81:
This updates repo deps for August 2024 for all deps except:
* `@bufbuild/protobuf`
* `@bufbuild/protoc-gen-es`
* `@bufbuild/protoplugin`

This is because the above deps have a released v2, but the currently released `@connectrpc` packages are not yet updated to accept v2 as a peer dep, so updating the above packages to their latest would result in peer dependency errors.

Also, the Svelte project's update of eslint caused errors because of its usage of `.eslintrc.cjs`. The instructions [here](https://eslint.org/docs/latest/use/configure/migration-guide#migrate-your-config-file) were followed to migrate that project's config.


svelte/src/routes/universal-ssr/+page.svelte
```diff
@@ -14,7 +14,6 @@
   console.log("universal-ssr +page.svelte", data);
 
   // Note that the full SayResponse Message type was able to pass through the SSR boundary in universal-ssr mode
-  data.response;
```
smaye81, line 17: This began failing eslint with:

```
  17:3  error  Expected an assignment or function call and instead saw an expression  @typescript-eslint/no-unused-expressions
```

timostamm: It was there for the comment. There's a near 100% chance frontend developers use vscode (or maybe webstorm), and they show the type right in the editor.

I'm fine with the comment next to the console.log call, did the same in v2 https://github.com/connectrpc/examples-es/blob/v2/svelte/src/routes/universal-ssr/%2Bpage.svelte#L13-L14

timostamm (approved):

PR #1974, opened 2024-10-31
V2: Update Angular example to latest Connect version

timostamm:


angular/src/connect/observable-client.ts
```diff
@@ -25,7 +25,7 @@ export function createObservableClient<T extends DescService>(
     service: T,
     transport: Transport
 ) {
-    return makeAnyClient(service, (method) => {
+    return makeAnyClient(service, (method: DescMethodUnary | DescMethodStreaming) => {
```
timostamm, line 28: This explicit type annotation is necessary to give the compiler the information to narrow down to a DescMethodUnary, which createUnaryFn needs.

This is not ideal. We should update makeAnyClient to have this type information in it's signature.

smaye81 (approved):

PR #2049, opened 2024-11-01
Update custom-client example

timostamm:


custom-client/src/strict-client.ts
```diff
@@ -1,70 +1,37 @@
...
 import type { Transport, CallOptions } from "@connectrpc/connect";
-import { createPromiseClient } from "@connectrpc/connect";
+import { createClient } from "@connectrpc/connect";
+import { OmitTypeName } from "./omit-type-name";
 
-// Tweak this to change the request signature like changing from PlainMessage to PartialMessage.
-type Request<R, I extends Message<I>> =
-  | Strict<R, PlainMessage<I>>
-  | StrictMessage<R, I>;
+type Request<I extends DescMessage> = OmitTypeName<MessageShape<I>>;
```
timostamm, line 14: This has become much more simple. The protobuf-es v2 types already are more type-safe than v1. We're simply omitting the `$typeName` property from the message types.

smaye81 (approved):

PR #2051, opened 2024-11-01
Update deps in bundle-size

timostamm:


bundle-size/parcel/package.json
```diff
@@ -28,13 +28,11 @@
     }
   },
   "dependencies": {
-    "@bufbuild/protobuf": "^1.10.0",
```
timostamm, line 31: Not sure what happened here.

smaye81 (approved):

PR #2055, opened 2024-11-01
Delete example with Create React App

timostamm:
The example scaffolded with create-react-app has dependencies with security vulnerabilities that cannot be resolved. We can try to re-create the example with a more recent version, but considering that we have many other examples using react, this doesn't seem very important. The official React docs use an online IDE nowadays.


smaye81 (approved):
Should delete this also: https://github.com/connectrpc/examples-es/blob/main/README.md?plain=1#L35

PR #2058, opened 2024-11-01
Delete examples with yarn

timostamm:
To ease maintenance a bit, this removes the two examples with yarn from this repository. 

We've never seen any issues with yarn the only gotcha is generating code with protoc outside of an npm script, which is document in https://www.npmjs.com/package/@bufbuild/protoc-gen-es


smaye81 (approved):
Same comment as CRA (remove from main README)

PR #2133, opened 2024-11-05
Fix React Native subpath imports

smaye81:
Using the Metro bundler with React Native requires enabling package exports ([docs](https://metrobundler.dev/docs/package-exports/)) via a Metro config file. This adds said file and some necessary dependencies.

Note that we also spell this out in our Protobuf-ES [docs](https://github.com/bufbuild/protobuf-es/blob/main/MANUAL.md#parcelmetro-fails-to-resolve-imports).

It seems odd to have to use the `react-native-svg-transformer` here. But without it, we get asset errors. This combination here appears to be the right incantation to get things to work. Taken from this [issue](https://github.com/facebook/react-native/issues/38282).

### DIFF
--- a/react-native/metro.config.js
+++ b/react-native/metro.config.js
@@ -12,8 +12,8 @@ const config = {
   },
   resolver: {
     unstable_enablePackageExports: true,
-    assetExts: assetExts.filter((ext) => ext !== "svg"),
-    sourceExts: [...sourceExts, "svg"],
+    // assetExts: assetExts.filter((ext) => ext !== "svg"),
+    // sourceExts: [...sourceExts, "svg"],
   },
 };
 
--- a/react-native/metro.config.js
+++ b/react-native/metro.config.js
@@ -1,9 +1,6 @@
 const { getDefaultConfig, mergeConfig } = require("@react-native/metro-config");
 
 const defaultConfig = getDefaultConfig(__dirname);
-const {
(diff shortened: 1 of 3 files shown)


react-native/metro.config.js
```diff
@@ -0,0 +1,15 @@
...
+
+const defaultConfig = getDefaultConfig(__dirname);
+
+const config = {
+  transformer: {
+    babelTransformerPath: require.resolve("react-native-svg-transformer"),
+    unstable_allowRequireContext: true,
+  },
+  resolver: {
+    unstable_enablePackageExports: true,
```
timostamm, line 11: It's added to the FAQ here: https://github.com/connectrpc/connectrpc.com/pull/218

If that's too difficult to link to, can we add a comment that replicates the gist of it? E.g.

> Connect-ES and Protobuf-ES use [package exports](https://nodejs.org/docs/latest-v12.x/api/packages.html#packages_exports). We need to enable support for them in Metro. See https://metrobundler.dev/docs/package-exports/

smaye81: Lmk if [this](https://github.com/connectrpc/examples-es/pull/2133/commits/81934b0dd17af45c8626d54f1aeaeedc261acfe2) is what you had in mind.

timostamm (approved):

PR #2241, opened 2024-12-12
Upgrade to React Native 0.76 and Expo 52

### LINKED ISSUE #2134: Upgrade to React-Native v0.76.0
React-Native v0.76 has been [released](https://reactnative.dev/blog/2024/10/23/release-0.76-new-architecture), but Expo [does not support](https://reactnative.dev/blog/2024/10/23/release-0.76-new-architecture#upgrade-to-076) it until SDK 52, which is not yet released. Once it is, we should upgrade the example.

smaye81:
Fixes #2134 

React Native 0.76 / Expo 52 have added full support for streaming. This updates our example to leverage this new functionality and remove notices/warnings that streaming is not yet supported.

```bash
$ npx expo-doctor                                                                                                                                                                        Enabled experimental React Native Directory checks. Unset the EXPO_DOCTOR_ENABLE_DIRECTORY_CHECK environment variable to disable this check.
✔ Check package.json for common issues
✔ Check Expo config for common issues
✔ Check for common project setup issues
✔ Check dependencies for packages that should not be installed directly
✔ Check for issues with Metro config
✔ Check npm/ yarn versions
✔ Check if the project meets version requirements for submission to app stores
✔ Validate packages against React Native Directory package metadata
✔ Check for app config fields that may not be synced in a non-CNG project
✔ Check Expo config (app.json/ app.config.js) schema
✔ Check that native modules do not use incompatible support packages
✔ Check for legacy global CLI installed locally
✔ Check native tooling versions
✔ Check that packages match versions required by installed Expo SDK
✔ Check that native modules use compatible support package versions for installed Expo SDK

Didn't find any issues with the project!
```

## Web
image

## Expo Go
image

## iOS Simulator
image


react-native/app/index.tsx
```diff
@@ -9,38 +10,57 @@ import {
...
+        // return new Promise<Response>((resolve) => {
+        //   fetch(input.toString(), {
+        //     ...init,
+        //     body: init?.body ?? undefined,
+        //     credentials: init?.credentials ?? undefined,
+        //     signal: init?.signal ?? undefined,
+        //   }).then((resp) => {
+        //     resolve(new Response(resp.body));
+        //   });
+        // });
```
timostamm, line 62: expo/fetch brings it's own types, even though expo/tsconfig.base includes the DOM lib. Little we can do except cast (we should actually test for input to be a string though).


```suggestion
      // Customize fetch with the Expo fetch implementation
      fetch: (input, init) => {
        if (typeof input !== "string") {
          throw new Error("expo/fetch requires the first argument to be a string URL");
        }
        return fetch(input, {
          ...init,
          body: init?.body ?? undefined,
          credentials: init?.credentials ?? undefined,
          signal: init?.signal ?? undefined,
        }) as unknown as Promise<Response>;
```

react-native/app/index.tsx
```diff
@@ -9,38 +10,57 @@ import {
...
-  const [responses, setResponses] = useState<Response[]>([
+  const [responses, setResponses] = useState<ConvoResponse[]>([
     {
       text: "What is your name?",
       sender: "eliza",
     },
   ]);
 
-  // Make the Eliza Service client
+  const abort = new AbortController();
```
timostamm, line 35: Not sure it's worth checking for this issue https://github.com/connectrpc/examples-es/issues/2134#issuecomment-2527886577. It has already been filed with a repro in https://github.com/expo/expo/issues/33553, and it looks like a fix is underway. May be worth calling it out in a comment here, maybe the fix doesn't land for a while.

react-native/app/index.tsx
```diff
@@ -9,38 +10,57 @@ import {
...
+
+  // Make the Eliza Service client using grpc-web transport.
```
timostamm, line 37: We're actually using the connect transport. Would be great to smoke test both with JSON and binary.

We also need to update README.md.

And https://connectrpc.com/docs/web/supported-browsers-and-frameworks

timostamm (commented):
Have been waiting years for this 🎉

react-native/app/index.tsx
smaye81: Yep, just wanted you to see how I was testing it. I was going to remove it when ready for review. Will add a link to the code here instead.

react-native/app/index.tsx
smaye81: Tested with binary format also and can confirm all 3 areas work (web, simulator, expo go).

Updated README and also created https://github.com/connectrpc/connectrpc.com/pull/244 for the docs.

react-native/app/index.tsx
```diff
@@ -9,38 +10,47 @@ import {
...
+          throw new Error(
+            "expo/fetch requires the first argument to be a string URL",
+          );
+        }
+        return fetch(input, {
+          ...init,
+          body: init?.body ?? undefined,
+          credentials: init?.credentials ?? undefined,
+          signal: init?.signal ?? undefined,
+        }) as unknown as Promise<Response>;
```
timostamm, line 52: Just realized - isn't this the same?

```suggestion
        return fetch(input, init) as unknown as Promise<Response>;
```

If this raises a type error, it'll most likely be fine to cast as RequestInit, similar to the response.

react-native/app/index.tsx
```diff
@@ -73,11 +86,7 @@ function Index() {
       } catch (err) {
         let text = "";
         if (err instanceof ConnectError) {
-          if (err.code === Code.Unimplemented) {
-            text = `Hi, ${statement}.  Streaming is not supported in React Native currently.`;
-          } else {
-            text = `A Connect error has occurred: ${err.code} - ${err.message}`;
-          }
+          text = `A Connect error has occurred: ${err.code} - ${err.message}`;
```
timostamm, line 89: A ConnectError's message includes the code:

```suggestion
          text = `A Connect error has occurred: ${err.message}`;
```

Can remove the import for `Code` to fix the linter warning.

timostamm (approved):

PR #2246, opened 2024-12-16
Upgrade Angular example to use standalone modules

### LINKED ISSUE #2238: Switch component to standalone in Angular example
The example in the `angular/` directory uses a non-standalone `NgModule`, but as of [Angular v19](https://blog.angular.dev/meet-angular-v19-7b29dfd05b84), standalone components are the default and recommended way to structure things. We should update the example to use the suggested approach.

smaye81:
Fixes #2238

As of [Angular v19](https://blog.angular.dev/meet-angular-v19-7b29dfd05b84), standalone components are the default. This upgrades the example to use standalone components. In addition, it also turns on the `strictStandalone` compiler option for Angular.

This also updates the test to use `TestBed.inject` rather than `TestBed.get` since the latter has been [deprecated](https://v17.angular.io/guide/deprecations#testing) since v9. This provides us a bit better type-safety when injecting providers.


timostamm (commented):
Nice, but we're not following latest best practices yet:

> You remove the concept of NgModule, but replace it with manual imports.

[Source](https://blog.angular-university.io/angular-standalone-components/).

The guide on NgModules gives clear guidance:

> IMPORTANT: The Angular team recommends using [standalone components](https://angular.dev/guide/components/anatomy-of-components#-imports-in-the-component-decorator) instead of NgModule for all new code.

We should remove app.module.ts. I gave it a stab in https://github.com/connectrpc/examples-es/pull/2247.

timostamm (approved):

PR #2247, opened 2024-12-17
Remove ng modules

timostamm:
Based on https://github.com/connectrpc/examples-es/pull/2246.


angular/src/app/app.component.ts
```diff
@@ -1,24 +1,21 @@
...
 @Component({
-  imports: [CommonModule, FormsModule],
   selector: "app-root",
-  templateUrl: "./app.component.html",
+  imports: [CommonModule, FormsModule],
   styleUrls: ["./app.component.css"],
+  templateUrl: "./app.component.html",
 })
 export class AppComponent {
+  client = inject(ELIZA);
```
timostamm, line 18: Instead of `@Inject`, we use the function `inject`.

angular/src/connect/connect.module.ts
```diff
@@ -1,29 +1,63 @@
...
+export const TRANSPORT = new InjectionToken<Transport>("connect.transport");
+
+export const INTERCEPTORS = new InjectionToken<Interceptor[]>(
+  "connect.interceptors",
+  {
+    factory: () => [],
+  },
+);
+
+export function createClientToken<T extends DescService>(service: T): InjectionToken<ObservableClient<T>> {
```
timostamm, line 16: This just makes it a bit easier to set up clients for many services.

timostamm (commented):
Haven't updated the tests.

angular/src/main.ts
```diff
@@ -1,12 +1,18 @@
-import { enableProdMode } from '@angular/core';
-import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
+/*
+ *  Protractor support is deprecated in Angular.
+ *  Protractor is used in this example for compatibility with Angular documentation tools.
```
smaye81, line 3: I'm not sure we should add a deprecated feature if we're trying to adhere to best practices in the example. If we're worried about users being confused looking at the example, can we just add a simple comment and not include the import and provider?

angular/src/connect/tokens.ts
```diff
@@ -1,17 +1,4 @@
...
-  "connect.interceptors",
-  {
-    factory: () => [],
-  },
-);
-
-export const ELIZA = new InjectionToken<ObservableClient<typeof ElizaService>>(
-  ElizaService.name,
-);
+export const ELIZA = createClientToken(ElizaService);
```
smaye81, line 4: Can we move this to `connect.module.ts`? Feels odd to have it it in its own `tokens` file now.

angular/src/main.ts
timostamm: This was blindly copied from https://angular.dev/tutorials/first-app 
Yes, we should remove it.

This also explains why the tutorial had an unused app.config.ts file. We can put the provider setup there. Done in https://github.com/connectrpc/examples-es/pull/2247/commits/83a1f0e9f5d9bc0b94534e65d32b90b794ca1f6f

angular/src/connect/connect.module.ts
```diff
@@ -1,29 +1,56 @@
...
-              interceptors: interceptors,
-            }),
-          deps: [INTERCEPTORS],
-        },
-      ],
-    };
-  }
+const TRANSPORT = new InjectionToken<Transport>("connect.transport");
+
+const INTERCEPTORS = new InjectionToken<Interceptor[]>("connect.interceptors", {
```
timostamm, line 12: Users can't add an interceptor if you don't export this const.

smaye81 (approved):

PR #2342, opened 2025-01-02
Update deps for Jan 2025

smaye81:
This updates the deps for Jan 2025.


paul-sachs:
Out of curiosity, why the downgrade from vite 6 to 5?

smaye81:
Huh. That's weird. Our upgrade script downgraded it for some reason. Not sure how that happened. I fixed it to be v6 again. I'll look into why that happened.

paul-sachs (approved):

PR #2343, opened 2025-01-03
Update license file

smaye81:
This updates the LICENSE file for 2025.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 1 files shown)


smaye81:
@timostamm we don't currently use `license-header` for the source files in this project. Not sure if we should or not since these are just examples.

timostamm (approved):
> @timostamm we don't currently use license-header for the source files in this project. Not sure if we should or not since these are just examples.

Many files have been scaffolded by frameworks. I prefer not to add a header to each file.

PR #2755, opened 2025-05-01
Update deps for May 2025

smaye81:
This updates the deps in the repo for the May dependabot run. A few notes:

* ESLint has been updated to the flat config for Next.js and Vue. 
* This does not include React Native updates, which are a bit more involved and will be done in a separate PR.


srikrsna-buf (approved):

vue/src/gen/connectrpc/eliza/v1/eliza_pb.ts
```diff
@@ -14,7 +14,7 @@
 
 // @generated by protoc-gen-es v2.2.5 with parameter "target=ts"
 // @generated from file connectrpc/eliza/v1/eliza.proto (package connectrpc.eliza.v1, syntax proto3)
-/* eslint-disable */
+
```
timostamm, line 17: Took me a second to see that we run "eslint . --fix", which will remove this unused directive.

Looks like it's time to add a plugin option for this annotation, similar to the `ts_nocheck` option.

timostamm (approved):

PR #2760, opened 2025-05-02
Update React Native and deps

smaye81:
This updates the main dependencies for the React Native example (i.e. React, React Native, Expo, etc.).

Since these deps are notoriously hard to incrementally upgrade, a new React project was generated using `npx create-expo-app react-native --template`. Then, code from the old react-native project was copied over and additional dependencies added (i.e. connect, buf, etc.).

Verified this setup works with the following:
* `npm run web` (in browser)
* `npm start` (via Expo Go on device)
* `npm run ios` (via iOS simulator and v17.5)


react-native/metro.config.js
```diff
@@ -4,10 +4,6 @@ const { mergeConfig } = require("metro-config");
 const defaultConfig = getDefaultConfig(__dirname);
 
 const config = {
-  transformer: {
-    babelTransformerPath: require.resolve("react-native-svg-transformer"),
-    unstable_allowRequireContext: true,
-  },
   resolver: {
```
smaye81, line 7: The app actually works without this entire file, but metro still lists package exports as [experimental](https://metrobundler.dev/docs/package-exports/), so probably best to keep it? 🤷‍♂️

timostamm: I suppose we don't need https://www.npmjs.com/package/react-native-svg-transformer because we don't import SVG files?

Expo 53: The package.json:exports field is now enabled by default in Metro bundler
https://expo.dev/changelog/sdk-53#the-packagejsonexports-field-is-now-enabled-by-default-in-metro-bundler

Below on line 12, we're still setting `unstable_enablePackageExports: true`. Does everything still work if you comment out the line?

If yes, please add an explanation with a link to the changelog. If not, let's keep it as is.

smaye81: Yeah, it still works. Removed the line and added a comment with link.

timostamm (approved):

Issue #2459, opened 2025-02-28
Update Angular example to use target=ts


danielalexis:
Currently it does not use target=ts, from a preliminary test it seems to work with target=ts, this is probably some old configuration.

PR #2762, opened 2025-05-05
Update buf.gen.yaml files to v2

smaye81:
This updates all the `buf.gen.yaml` files used in our examples to use the v2 layout.

In addition, it adds a `- target=ts` opt to angular, parcel, and react native. Previously, these had comments saying that output wasn't supported, but this now passes CI and these apps run fine using this opt.


bundle-size/esbuild/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "consumer-esbuild",
   "scripts": {
     "pregenerate": "rm -rf dist",
-    "generate": "npx buf generate ../proto --template ../buf.gen.yaml",
+    "generate": "buf generate ../proto --template ../buf.gen.yaml",
```
smaye81, line 5: Left in the `../proto` for the bundle size examples because putting `../proto` as an input into the `buf.gen.yaml` is confusing.

timostamm (commented):
So we can close https://github.com/connectrpc/examples-es/issues/2459 with this?

Looks like `target=ts` is also new for react-native, and needs a re-gen.

timostamm (approved):

timostamm (approved):

PR #2851, opened 2025-08-01
Update deps for August 2025

smaye81:
SSIA


bundle-size/README.md
```diff
@@ -17,16 +17,16 @@ The results are:
 <!--- RESULTS-START -->
 | Name | Size | Artifact |
 |------|-----:|----------|
-| esbuild (js) | 123,964 bytes | dist/js/index.js |
-| esbuild (ts) | 123,964 bytes | dist/ts/index.js |
+| esbuild (js) | 135,295 bytes | dist/js/index.js |
```
paul-sachs, line 20: Hmm, i wonder what caused an additional 10k+ increase to bundle size 🤔

smaye81: Yeah, Parcel didn't change. I'm wondering if it's something in our configuration for these bundlers? @timostamm any ideas what may be the culprit here?

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "examples-es",
```
timostamm, line 2: This file shouldn't be here.

bundle-size/README.md
timostamm: Protobuf-ES 2.6.1 includes an update to google-protobuf v31.1 with a small bump in bundle size (see [PR](https://github.com/bufbuild/protobuf-es/pull/1168)), but that should be _much_ smaller than 10kb.

Would be good to know what's going on here. Can you diff the bundled output against the previous one?

PR #2852, opened 2025-08-04
Fix parcel bundle size report

timostamm:


bundle-size/parcel/package.json
```diff
@@ -3,7 +3,9 @@
   "scripts": {
     "pregenerate": "rm -rf dist",
     "generate": "buf generate ../proto --template ../buf.gen.yaml",
-    "build": "parcel build --log-level error --no-cache --no-content-hash",
+    "build": "npm run build:js && npm run build:ts",
+    "build:js": "parcel build --log-level error --no-cache --no-content-hash --target=js",
+    "build:ts": "parcel build --log-level error --no-cache --no-content-hash --target=ts",
```
timostamm, line 8: We have to run parcel twice, so that it produces two separate and independent outputs. Otherwise, it still generates imports, and we didn't account for those imports when summing up the size.

Yaheo (changes requested):
M

PR #3062, opened 2026-05-07
Remove Dependabot

ajeetdsouza:
Remove Dependabot; this repo contains example code and does not need frequent dependency updates.


timostamm (approved):
> this repo contains example code and does not need frequent dependency updates

Agreed 👍

